PR #3, opened 2018-03-16
Added openjdk build

MaSven:
As discussed in [issue 629](https://github.com/sdkman/sdkman-cli/issues/629) i added the adoptopenjdk builds. But only the latest ones.


MaSven:
The strange version numbers coming from the build system. Maybe they will get consistent over time. I also had to add something because for example the java 9 build for hotspot and openj9 have the same buildnumber. That is why there is for example jk9 and jkd9j9.

marc0der:
@MaSven thanks so much for the excellent work. I've made a feel tweaks, in particular to the version strings so that they align with the rest of our jdk versions. I've also removed the j9 jdks for now until we have a demand for them.

PR #5, opened 2018-03-21
Change the openjdk to adoptjdk

MaSven:
I think people will get confused with the openjdk suffic since openjdk releases it's own builds. So i changed the suffix for the adoptopenjdk to -adoptopenjdk. Also to make clear that this are not the openjdk builds.

### DIFF
--- a/src/main/scala/io/sdkman/changelogs/Migrations.scala
+++ b/src/main/scala/io/sdkman/changelogs/Migrations.scala
@@ -41,7 +41,6 @@ class Migrations {
     updateCandidateDefault("maven", "3.5.3")
   }
 
-
   @ChangeSet(order = "008", id = "008-change_openjdk_to_adoptopenjdk", author = "MaSven")
   def migrate008(implicit db: MongoDatabase) = {
     updateVersion("9u181-openjdk", "9u181-adoptopenjdk")
@@ -53,6 +52,15 @@ class Migrations {
   def migrate007(implicit db: MongoDatabase) = {
     insertVersions(CandidateVersion("scala", "2.12.5", Some("UNIVERSAL"), "https://downloads.lightbend.com/scala/2.12.5/scala-2.12.5.zip"))
     updateCandidateDefault("scala", "2.12.5")
+
+  }
+
+  @ChangeSet(order = "009", id = "009-add_java_10_openjdk", author = "MaSven")
+  def migrate009(implicit db: MongoDatabase) = {
+    insertVersions(
+      CandidateVersion("java", "10-openjdk", Some("WINDOWS_64"), "https://download.java.net/java/GA/jdk10/10/binaries/openjdk-10_windows-x64_bin.tar.gz"),
+      CandidateVersion("java", "10-openjdk", Some("MAC_OSX"), "https://download.java.net/java/GA/jdk10/10/binaries/openjdk-10_osx-x64_bin.tar.gz"),
+      CandidateVersion("java", "10-openjdk", Some("LINUX_64"), "https://download.java.net/java/GA/jdk10/10/binaries/openjdk-10_linux-x64_bin.tar.gz"))
 
   }
 }
--- a/src/main/scala/io/sdkman/changelogs/Migrations.scala
+++ b/src/main/scala/io/sdkman/changelogs/Migrations.scala
@@ -1,10 +1,11 @@
 package io.sdkman.changelogs
 
-import com.github.mongobee.changeset.{ChangeLog, ChangeSet}
+import com.github.mongobee.changeset.{ ChangeLog, ChangeSet }
 import com.mongodb.client.MongoDatabase
 import org.bson.Document
 
 import scala.collection.JavaConverters._
+import scala.collection.mutable.ListBuffer
 
 @ChangeLog(order = "001")
 class VersionsMigration {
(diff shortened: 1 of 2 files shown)


marc0der:
Not certain I agree. AdoptOpenJDK are actually the official openjdk builds now, and the ones that you refer to are merely what debian/fedora etc package themselves. I'd prefer rolling with openjdk, because this is precisely what it is in it's purest form.

MaSven:
ok so i was confused because here http://jdk.java.net/10/ java 10 is build for all major platforms but in adoptopenjdk https://adoptopenjdk.net/releases.html?variant=openjdk10 only linux and mac is released.

src/main/scala/io/sdkman/changelogs/Migrations.scala
```diff
@@ -35,16 +35,32 @@ class Migrations {
...
     updateCandidateDefault("scala", "2.12.5")
+
+  }
+
+  @ChangeSet(order = "009", id = "009-add_java_10_openjdk", author = "MaSven")
+  def migrate009(implicit db: MongoDatabase) = {
+    insertVersions(
+      CandidateVersion("java", "10-openjdk", Some("WINDOWS_64"), "https://download.java.net/java/GA/jdk10/10/binaries/openjdk-10_windows-x64_bin.tar.gz"),
+      CandidateVersion("java", "10-openjdk", Some("MAC_OSX"), "https://download.java.net/java/GA/jdk10/10/binaries/openjdk-10_osx-x64_bin.tar.gz"),
+      CandidateVersion("java", "10-openjdk", Some("LINUX_64"), "https://download.java.net/java/GA/jdk10/10/binaries/openjdk-10_linux-x64_bin.tar.gz"))
```
marc0der: These are actually Oracle's openjdk distributions. I think adopt are considered more official by now.

src/main/scala/io/sdkman/changelogs/Migrations.scala
```diff
@@ -35,16 +35,32 @@ class Migrations {
...
-
   @ChangeSet(order = "006", id = "003-add_maven_353", author = "hho")
   def migrate006(implicit db: MongoDatabase) = {
     insertVersions(CandidateVersion("maven", "3.5.3", Some("UNIVERSAL"), "https://archive.apache.org/dist/maven/maven-3/3.5.3/binaries/apache-maven-3.5.3-bin.zip"))
     updateCandidateDefault("maven", "3.5.3")
   }
 
+  @ChangeSet(order = "008", id = "008-change_openjdk_to_adoptopenjdk", author = "MaSven")
+  def migrate008(implicit db: MongoDatabase) = {
+    updateVersion("9u181-openjdk", "9u181-adoptopenjdk")
```
marc0der: This name is too verbose, and takes up too much real estate in the list view. I think we should stick with openjdk for now.

src/main/scala/io/sdkman/changelogs/Migrations.scala
```diff
@@ -1,6 +1,6 @@
 package io.sdkman.changelogs
 
-import com.github.mongobee.changeset.{ChangeLog, ChangeSet}
+import com.github.mongobee.changeset.{ ChangeLog, ChangeSet }
```
marc0der: Not related to the PR, and not the standard scala code style.

src/main/scala/io/sdkman/changelogs/package.scala
```diff
@@ -1,26 +1,29 @@
...
 
-  case class CandidateVersion(candidate: String,
-                              version: String,
-                              platform: Option[String],
-                              url: String)
+  case class CandidateVersion(
+    candidate: String,
+    version:   String,
+    platform:  Option[String],
+    url:       String)
```
marc0der: Reformatting not a part of this PR. Also, once again not standard scala code style.

src/main/scala/io/sdkman/changelogs/package.scala
```diff
@@ -44,4 +47,9 @@ package object changelogs {
 
   def updateCandidateDefault(candidate: String, version: String)(implicit db: MongoDatabase): Document =
     db.getCollection("candidates").findOneAndUpdate(Filters.eq("candidate", candidate), Updates.set("default", version))
+
+  def updateVersion(version: String, newVersion: String)(implicit db: MongoDatabase): Unit = {
+    db.getCollection("versions").updateMany(Filters.eq("version", version), Updates.set("version", newVersion))
```
marc0der: `updateMany` could potentially be dangerous as version could be very ambiguous. I like the idea of adding this helper function though...

marc0der:
@MaSven It's late here now, let me sleep on this a bit. I think adoptjdk does sound kinda cool ;-)

src/main/scala/io/sdkman/changelogs/Migrations.scala
MaSven: Sorry not a scala dev and just downloaded the scala ide for the pull requests. Do you have some form of setting for the scala ide so i have the right formatting?

src/main/scala/io/sdkman/changelogs/package.scala
MaSven: Ok so i think a better way would be to find them one by one and after that update the version?

MaSven:
yes and it is shorter. Makes less typing.

src/main/scala/io/sdkman/changelogs/Migrations.scala
MaSven: This wasn't intended for this PR. I pushed to the wrong branch here sorry for that. I wanted to make a new PR for this one.

marc0der:
After the [Twitter poll](https://twitter.com/sdkman_/status/977076310571978757) on Friday it is conclusive that everyone _wants_ the openjdk from java.net. This will become our official openjdk distribution of choice and all others will be removed soon (no point in hosting multiple openjdk distros as they all contain the same). As a result of this, I'm closing this PR for now.

tony--:
Any chance of revisiting this, @marc0der @MaSven ?
I plan to use adoptopenjdk because of the longer support life.

PR #11, opened 2018-04-23
Add GraalVM migrations

shelajev:
I've added the GraalVM CE migrations to make GraalVM available on the sdkman. I don't know how to test sdkman agains the local mongo, but the migrations against it run okay.

Currently, there binaries for GraalVM CE are only for Linux.


marc0der:
@shelajev thanks for your PR to add GraalVM to sdkman! I just tried it out and in testing noticed a caveat. I'll take the discussion back to https://github.com/oracle/graal/issues/367 where this discussion was started.

PR #19, opened 2018-06-10
Add bpipe 0.9.9.5 to SDKMan

ssadedin:
This PR is to add Bpipe, a tool that I am the author of (see http://bpipe.org) to SDKMan.

Thanks!


ssadedin:
Ok - I've taken the version out. Not sure if it is a problem that the Candidate still refers to a default version, even though it doesn't exist?

Thanks!

marc0der:
Yes that will break, simply turn the default version to None or remove it
all together (it will automatically be set to None)

ssadedin:
Ok - removed the version for now.

Let me know any other issues!

marc0der:
Thanks!

PR #24, opened 2018-08-10
Add Java 11 and 12 EA candidates

mdeinum:
With this commit the EA candidates for Java 11 and 12 are added. 
Java 11 both has an Oracle and OpenJDK version, the Java 12 version
only has an OpenJDK version. 

The versioning is like *11.ea25-openjdk* as using *11.ea+25-openjdk* would 
result in a version longer then 15 characters. When the EA candidates can
be removed we can revert back to *11.0.0-openjdk* again.

### DIFF
--- a/src/main/scala/io/sdkman/changelogs/JavaMigrations.scala
+++ b/src/main/scala/io/sdkman/changelogs/JavaMigrations.scala
@@ -309,4 +309,24 @@ class JavaMigrations {
       .validate()
       .insert()
   }
+
+  @ChangeSet(order = "033", id = "033-add_openjdk_java_11-ea-25", author = "mdeinum")
+  def migrate033(implicit db: MongoDatabase) = {
+    List(
+      Version("java", "11.e.25-openjdk", "https://download.java.net/java/early_access/jdk11/25/GPL/openjdk-11-ea+25_linux-x64_bin.tar.gz", Linux64),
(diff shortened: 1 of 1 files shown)


marc0der:
I'd like to propose two changes.

Firstly, I would prefer if we drop the Oracle JDK and just go with OpenJDK for EA releases.

Secondly I want to propose that we make the numbering scheme `11.ea.25-openjdk` so that we don't break the sorting algorithm.

marc0der:
Sorry, and another thing is that all the binaries need to be tarballs if possible. Have they changed their openjdk binary format [since JDK 10](https://github.com/sdkman/sdkman-db-migrations/blob/master/src/main/scala/io/sdkman/changelogs/JavaMigrations.scala#L12)?

marc0der:
Just saw your comment about the version truncation. Maybe we should try `11.e.25-openjdk` as the delimiters are very important. wdyt?

mdeinum:
-Dropped Oracle JDK
- Changed versioning scheme. 

Regarding the binaries, I followed the download links for the respective versions. Only linux and Mac have tarballs, windows has a zip. This looks consistent to java 9 and 10. (The Oracle JDK was different indeed but that is removed now).

marc0der:
I just checked, and it looks like the format of the openjdk has indeed changed from JDK [9/10](http://jdk.java.net/10) to [11/12](http://jdk.java.net/11).

That said, I've released a [change in the backend](https://github.com/sdkman/sdkman-hooks/commit/72d3e8f4af7d6bfbaaf3555888468f8e4bd7892d) that will allow for this format to work.

marc0der:
BTW, thanks for your contribution!

PR #25, opened 2018-08-13
Update Zulu JDKs

vpavic:
This PR updates Zulu JDKs to current versions. There's also an additional commit that adds Zulu `11-ea`, if that's acceptable addition.

### DIFF
--- a/src/main/scala/io/sdkman/changelogs/JavaMigrations.scala
+++ b/src/main/scala/io/sdkman/changelogs/JavaMigrations.scala
@@ -329,4 +329,43 @@ class JavaMigrations {
       .validate()
       .insert()
   }
+
+  @ChangeSet(order = "035", id = "035-add_zulu_6_0_113", author = "vpavic")
+  def migrate035(implicit db: MongoDatabase) = {
+    Seq(Linux64, Windows).foreach(platform => removeVersion(candidate = "java", version = "6.0.107-zulu", platform))
+    List(
+      Version("java", "6.0.113-zulu", "https://cdn.azul.com/zulu/bin/zulu6.21.0.3-jdk6.0.113-linux_x64.tar.gz", Linux64),
+      Version("java", "6.0.113-zulu", "https://cdn.azul.com/zulu/bin/zulu6.21.0.3-jdk6.0.113-win_x64.zip", Windows)
+    ).validate().insert()
+  }
+
+  @ChangeSet(order = "036", id = "036-add_zulu_7_0_191", author = "vpavic")
+  def migrate036(implicit db: MongoDatabase) = {
+    Seq(Linux64, Windows).foreach(platform => removeVersion(candidate = "java", version = "7.0.181-zulu", platform))
+    List(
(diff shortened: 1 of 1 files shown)


src/main/scala/io/sdkman/changelogs/JavaMigrations.scala
```diff
@@ -329,4 +329,53 @@ class JavaMigrations {
...
+      Version("java", "10.0.2-zulu", "https://cdn.azul.com/zulu/bin/zulu10.3+5-jdk10.0.2-macosx_x64.tar.gz", MacOSX)
+    ).validate().insert()
+  }
+
+  @ChangeSet(order = "039", id = "039-add_zulu_11-ea", author = "vpavic")
+  def migrate039(implicit db: MongoDatabase) = {
+    List(
+      Version("java", "11-ea-zulu", "https://cdn.azul.com/zulu/bin/zulu11.1+23-ea-jdk11-linux_x64.tar.gz", Linux64),
+      Version("java", "11-ea-zulu", "https://cdn.azul.com/zulu/bin/zulu11.1+23-ea-jdk11-win_x64.zip", Windows),
+      Version("java", "11-ea-zulu", "https://cdn.azul.com/zulu/bin/zulu11.1+23-ea-jdk11-macosx_x64.zip", MacOSX)
```
marc0der: I'd prefer not to be hosting too many EA releases for now. As we already have OpenJDK 11 & 12, would you mind if we drop 11 EA all together for Zulu?

vpavic:
Thanks for the feedback @marc0der - I've removed `7.0.191-zulu` for `MacOSX` and the second commit that contained `11-ea`.

src/main/scala/io/sdkman/changelogs/JavaMigrations.scala
```diff
@@ -329,4 +329,43 @@ class JavaMigrations {
...
+    Seq(Linux64, Windows).foreach(platform => removeVersion(candidate = "java", version = "6.0.107-zulu", platform))
+    List(
+      Version("java", "6.0.113-zulu", "https://cdn.azul.com/zulu/bin/zulu6.21.0.3-jdk6.0.113-linux_x64.tar.gz", Linux64),
+      Version("java", "6.0.113-zulu", "https://cdn.azul.com/zulu/bin/zulu6.21.0.3-jdk6.0.113-win_x64.zip", Windows)
+    ).validate().insert()
+  }
+
+  @ChangeSet(order = "036", id = "036-add_zulu_7_0_191", author = "vpavic")
+  def migrate036(implicit db: MongoDatabase) = {
+    Seq(Linux64, Windows, MacOSX).foreach(platform => removeVersion(candidate = "java", version = "7.0.181-zulu", platform))
```
marc0der: I think we should probably leave 7.0.181-zulu for OSX in tact as it still seems to be working as a tarball. We can replace it with 7.0.191 for the other platforms.

vpavic: Whoops - I missed that one, you're right, of course.

marc0der:
Thanks for your contrib @vpavic, always great work.

PR #31, opened 2018-09-17
Add julia 1.0.0

gitqlt:
the language they’ve created is called Julia


marc0der:
Unfortunately we only handle zip archives by default.

We can accommodate tarballs, but no exe and dmg files I'm afraid. These file formats require custom hooks to handle the contents of these files.

PR #36, opened 2018-10-06
Upgrade Java

renannprado:


renannprado:
Ah, I see.
It would be nice though if they provided the release candidate versions as well.

renannprado:
@marc0der I've removed the gradle migrations

marc0der:
Thanks for contributing!

marc0der:
@renannprado You might want to ask them on their own forums why they haven't been releasing Gradle 5.x builds. Maybe something not set up in their build pipeline?

renannprado:
@marc0der I'll certainly do that. Thanks for the hing!

PR #48, opened 2018-11-27
add Spark 2.1.3

machielg:
I also disabled the Java 11 ea migrations that are not working anymore because the URLs give 404


marc0der:
Sorry to be a pain but can we remove them instead of commenting them out?

machielg:
> Sorry to be a pain but can we remove them instead of commenting them out?

No problem, I removed them now

marc0der:
Bedankt jongen! 😀

PR #49, opened 2018-11-28
Add OpenJDK versions [8, 10, 11] with OpenJ9

munukutla:
Added migrations to use OpenJDK with OpenJ9.

| Version          | Platforms      |
|------------------|----------------|
| OpenJDK 8 (LTS)  | Linux, Windows |
| OpenJDK 10       | Linux, Windows |
| OpenJDK 11 (LTS) | Linux, Mac, Windows |

### DIFF
--- a/src/main/scala/io/sdkman/changelogs/JavaMigrations.scala
+++ b/src/main/scala/io/sdkman/changelogs/JavaMigrations.scala
@@ -322,4 +322,31 @@ class JavaMigrations {
     Seq(Linux64, MacOSX, Windows).foreach(removeVersion("java", "12.ea.17-open", _))
   }
 
+  @ChangeSet(order = "065", id = "065-add_eclipsej9_openjdk_8", author = "munukutla")
+  def migrate065(implicit db: MongoDatabase) = {
+    List(
+      Version("java", "8-j9", "https://github.com/AdoptOpenJDK/openjdk8-binaries/releases/download/jdk8u192-b12_openj9-0.11.0/OpenJDK8U-jdk_x64_linux_openj9_8u192b12_openj9-0.11.0.tar.gz", Linux64),
+      Version("java", "8-j9", "https://github.com/AdoptOpenJDK/openjdk8-binaries/releases/download/jdk8u192-b12_openj9-0.11.0/OpenJDK8U-jdk_x64_windows_openj9_8u192b12_openj9-0.11.0.zip", Windows))
+      .validate()
+      .insert()
+  }
+
+  @ChangeSet(order = "066", id = "066-add_eclipsej9_openjdk_10", author = "munukutla")
+  def migrate065(implicit db: MongoDatabase) = {
+    List(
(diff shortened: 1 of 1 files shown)


munukutla:
@marc0der Sorry missed notifying on Gitter. This was a response to https://github.com/eclipse/openj9/issues/1924. I took your comment https://github.com/eclipse/openj9/issues/1924#issuecomment-389866302 as an open invitation :stuck_out_tongue_winking_eye:

src/main/scala/io/sdkman/changelogs/JavaMigrations.scala
```diff
@@ -304,4 +304,32 @@ class JavaMigrations {
   def migrate065(implicit db: MongoDatabase) = {
     Seq(Linux64, Windows, MacOSX).foreach(platform => removeVersion(candidate = "java", version = "11.e.26-openjdk", platform))
   }
+
+  @ChangeSet(order = "066", id = "065-add_eclipsej9_openjdk_8", author = "munukutla")
+  def migrate066(implicit db: MongoDatabase) = {
+    List(
+      Version("java", "8-open-j9", "https://github.com/AdoptOpenJDK/openjdk8-binaries/releases/download/jdk8u192-b12_openj9-0.11.0/OpenJDK8U-jdk_x64_linux_openj9_8u192b12_openj9-0.11.0.tar.gz", Linux64),
```
marc0der: We will need to follow the existing naming convention for Java candidates. Currently, this comes in the form of `major.minor.patch-distro`. As an example, this one would be `8.0.192-adopt`. We have a hard limit of 15 characters per version number, the reason being that the console column view in `sdk ls java` will truncate the version string if longer. I would suggest the `distro` part of the version string to be something concise like `adopt`, as this is currently about the party vending the software.

munukutla: So can we ignore the fact that AdoptOpenJDK also has distributions with HotSpot VM? Because `8.0.192-adopt` is available with both J9 and Hotspot.

marc0der: Yes, it would be good if we can represent this somehow, but not sure how we can pull that off with the limited characters available. Unless we find a way to abbreviate the distro to capture both. Like `8.0.192-adptj9` or something similar.

munukutla: I can foresee a number of OpenJDK distributions being released in the next couple of years. Barring any coincidences, many of them might end up having longer distribution names, like Amazon Corretto etc. Can we look to extend the 15 character limit so that it doesn't pin our heels later?

OpenJDK + Hotspot => `8.0.192-openhtspt` (it's already weird)
OpenJDK + OpenJ9 => `8.0.192-openj9` (kinda obvious)

marc0der: We would need to cross that bridge when we get there. AFAIK openjk does not have any plans in shipping anything besides hotspot vms.

Changing the length is an expensive exercise because we would need to update the list view, the vendor publishing api as well as validations in the db migrations. For me this would be premature optimisation. Besides, 15 characters should be ample for version string.

Corretto will be a single distribution and will get the suffix `amzn` or something similar. We already have _all_ the current vendors represented including graal, openjdk, zulu and oracle. I'm confident we'll manage with the scheme we have.

In future, let's please try having these discussion in Gitter prior to raising a PR.

PR #52, opened 2019-01-09
Upgrade Java 12 and add Java 13

mdeinum:
With this commit we upgrade Java 12 to 12.ea.26 and
we introduce Java 13.ea.02 as a version to install.


marc0der:
Thanks for contributing!

PR #53, opened 2019-01-13
Add Java 8,11 with FX

renannprado:


renannprado:
hey @marc0der do you think it makes sense to add java-fx versions into the sdkman?
I needed to use FX these days and I really missed that it wasn't in sdkman :)

marc0der:
Thanks for the contribution!

PR #59, opened 2019-01-20
Removes offending migrations that break tests

rvprasad:


rvprasad:
I will resubmit a cleaner PR.

PR #64, opened 2019-01-28
Added Leiningen 2.8.3 to JavaMigrations changelog

simion-iulian:


marc0der:
Apart from the conflicts, we don't really want to be deleting a bunch of other migrations. At most, we could disable the validation on some of the defunct URLs in the `JavaMigrations`.

In terms of Leiningen 2.8.3, this binary is not compatible with SDKMAN in that it is not properly formed. It doesn't have an enclosing folder, nor does it have a bin folder that contains all executables. Version 2.7.1 had a well formed binary that was compliant with the standard.

simion-iulian:
True, couldn't get it to run locally without deleting the `zulu` links for quite a few JDK versions. It was throwing a `sun.security.validator.ValidatorException`.

PR #70, opened 2019-02-17
Adds kobalt as a SDK

rvprasad:


marc0der:
Kobalt was previously on SDKMAN but we decided to pull it due to inactivity and the fact that Gradle is now the preferred build tool for Kotlin.

For now at least we probably won't be adding it back.

PR #94, opened 2019-04-17
Remove Java 9 and 10 and old GraalVM

jorsol:
Java 9 and Java 10 (zulu and open variants) are now out of support, so they should not be available.

GraalVM < 1.0.0-rc-15-grl should also be removed, new versions are just added but not removed the previous one.


marc0der:
This is all based on your own assumptions and was never discussed on Gitter.

* JDK 9 and 10 _should_ remain available for historical purposes, even they are no longer officially supported.
* GraalVM is keeping it's history (for now) since the Micronaut team have explicitly asked for them to stay available. Micronaut only works with certain versions of GraalVM. @ilopmar will be able to advise which versions _can_ be deleted.

jorsol:
Sure, fair enough.

ilopmar:
As Graal RC16 will be released soon (it was tagged last week) I think we can remove from rc8 to rc12 
 both included.
Micronaut 1.1.0 publishes rc13 in the bom so we should keep that for a while.

PR #95, opened 2019-04-18
Add JDK's

philiplourandos:
* Added Zulu Java7 and Java8 jdk's
* Added Amazon Java 8 and Java 11 jdk's


marc0der:
Great! Thank you!

PR #111, opened 2019-05-13
Add Graal 19.0.0 and remove Graal RC14

ilopmar:
This PR add Graal 19.0.0 and removes old RC14. We'll keep RC15 and 16 for a while because the new one introduced some breaking changes.

### DIFF
--- a/src/main/scala/io/sdkman/changelogs/JavaMigrations.scala
+++ b/src/main/scala/io/sdkman/changelogs/JavaMigrations.scala
@@ -962,17 +962,17 @@ class JavaMigrations {
     List(
       Version(
         candidate = "java",
-        version = "19.0.0-grl-ce",
+        version = "19.0.0-grl",
         url = "https://github.com/oracle/graal/releases/download/vm-19.0.0/graalvm-ce-linux-amd64-19.0.0.tar.gz",
         platform = Linux64),
       Version(
         candidate = "java",
-        version = "19.0.0-grl-ce",
+        version = "19.0.0-grl",
         url = "https://github.com/oracle/graal/releases/download/vm-19.0.0/graalvm-ce-darwin-amd64-19.0.0.tar.gz",
         platform = MacOSX),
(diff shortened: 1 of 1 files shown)


src/main/scala/io/sdkman/changelogs/JavaMigrations.scala
```diff
@@ -957,4 +957,27 @@ class JavaMigrations {
     Seq(Linux64, MacOSX, Windows).foreach(removeVersion("java", "13.ea.19-open", _))
   }
 
+  @ChangeSet(order = "138", id = "138-add_graalvm_19_0_0", author = "ilopmar")
+  def migrate138(implicit db: MongoDatabase) = {
+    List(
+      Version(
+        candidate = "java",
+        version = "19.0.0-graal-ce",
```
marc0der: We'll need to stick to using the same `-grl` suffix as before in order for the _sdkman-hooks_ service to handle these versions.

ilopmar: Pushed the change. Feel free to squash the commits when merging them

marc0der: Thanks @ilopmar. You have some serious Scala skills there!!! :laughing:

marc0der: Sorry for being a pain, it needs to be `-grl` without the `-ce` for the hooks to be picked up.

ilopmar: I'm not so sure, 3 commits to make a simple change :(

marc0der: You did well, thanks for doing this! :metal:

marc0der:
Thank you!

ilopmar:
Thanks buddy!

PR #141, opened 2019-07-25
* Added Amazon Corretto 11.0.4 and 8.0.222 JDKs

philiplourandos:


philiplourandos:
> Zulu Java 7 update 232 is already included, and it's not from Amazon

didn't see that. backed out that commit

PR #143, opened 2019-07-26
Restore GraalVM 19.0.2

dsyer:
Minor versions are rare in a JDK but Graal has them and there are breaking changes in 19.1. This restores the latest 19.0 release.


src/main/scala/io/sdkman/changelogs/JavaMigrations.scala
```diff
@@ -548,4 +548,29 @@ class JavaMigrations {
       .insert()
     Seq(Linux64, MacOSX, Windows).foreach(removeVersion("java", "8.0.212.j9-adpt", _))
   }
+  
+   @ChangeSet(order = "178", id = "178-restore_graalvm_19_0", author = "dsyer")
+  def migrate164(implicit db: MongoDatabase) = {
```
marc0der: The method should match the changeset annotation number. Can you make that 164?

PR #147, opened 2019-08-21
Add GraalVM 19.2.0

eddumelendez:


src/main/scala/io/sdkman/changelogs/JavaMigrations.scala
```diff
@@ -579,4 +560,30 @@ class JavaMigrations {
...
+				vendor = Some(Graal)),
+			Version(
+				candidate = "java",
+				version = "19.2.0-grl",
+				url = "https://github.com/oracle/graal/releases/download/vm-19.2.0/graalvm-ce-windows-amd64-19.2.0.zip",
+				platform = Windows,
+				vendor = Some(Graal)))
+			.validate()
+			.insert()
+		Seq(Linux64, MacOSX).foreach(platform => removeVersion(candidate = "java", version = "19.1.1-grl", platform))
```
ilopmar: Please do not remove old versions, specially if they are from another minor version. Some tools and frameworks do not support the latest version and still need to use previous one.

src/main/scala/io/sdkman/changelogs/JavaMigrations.scala
```diff
@@ -385,25 +385,6 @@ class JavaMigrations {
     Seq(Linux64, Windows, MacOSX).foreach(platform => removeVersion(candidate = "java", version = "12.0.1-zulu", platform))
   }
 
-  @ChangeSet(order = "163", id = "163-add_graalvm_19_1_1", author = "wololock")
```
ilopmar: I'm not 100% sure but (@marc0der will confirm it) I think you shouldn't remove previous migrations. At least it's the first time I see removing an old migration when adding a new new.

marc0der: @ilopmar I've asked for the trimming down of obsolete changesets as the `JavaMigrations` file is getting very bloated.

I was also toying with the idea of deprecating `JavaMigrations` and branching out into every JDK distribution having it's own migration file. Then we could refrain from deleting changesets.

In this case, `19.1.1` is not obsolete though as it will still remain in production for a while. @eddumelendez can we keep for now?

jorsol: Oracle now has a [version roadmap](https://www.graalvm.org/docs/release-notes/version-roadmap/) for graalvm. It's clear that what should be a _minor_ version in semantic versioning, it's in fact a _major_ version for graalvm (they use a weird [calver](http://calver.org/) `YY.MAJOR.MICRO`).

So a proposal could be, keep the last 3 major versions plus the last LTS patch.

src/main/scala/io/sdkman/changelogs/JavaMigrations.scala
```diff
@@ -579,4 +560,30 @@ class JavaMigrations {
     Seq(Linux64, MacOSX, Windows).foreach(removeVersion("java", "14.ea.6-open", _))
   }
 
+	@ChangeSet(order = "182", id = "182-add_graalvm_19_2_0", author = "eddumelendez")
```
wololock: Please fix the indention of the added code, so it uses the same code style as other parts in the file. Thanks in advance!

marc0der: I hate to do it, but I might be adding [scalafmt](https://scalameta.org/scalafmt/) to this project soon.

src/main/scala/io/sdkman/changelogs/JavaMigrations.scala
eddumelendez: @ilopmar thanks for letting me know about it. that version is not going to be removed then

src/main/scala/io/sdkman/changelogs/JavaMigrations.scala
eddumelendez: I have applied the right format

eddumelendez:
PR updated

marc0der:
Thanks for contributing, and thanks for everyone's input!

PR #155, opened 2019-09-18
* Added zulu JDK 13.0.0

philiplourandos:


src/main/scala/io/sdkman/changelogs/JavaMigrations.scala
```diff
@@ -629,4 +629,13 @@ class JavaMigrations {
       .insert()
   }
 
+  @ChangeSet(order = "186", id = "186-zulu_13.0.0", author = "philiplourandos")
+  def migrate186(implicit db: MongoDatabase) = {
+    List(
+      Version("java", "13.0.0", "https://cdn.azul.com/zulu/bin/zulu13.27.9-ca-jdk13-linux_x64.tar.gz", Linux64, Some(Zulu)),
```
marc0der: The version should be `13.0.0-zulu`, as this will be used as the identifier.

philiplourandos: copy that

PR #156, opened 2019-09-19
Add OpenJDK 13 GA 33

javafx-dev:


javafx-dev:
paul@paul:~$ mkdir tmp
paul@paul:~$ cd tmp
paul@paul:~/tmp$ wget https://download.java.net/java/GA/jdk13/5b8a42f3905b406298b72d750b6919f6/33/GPL/openjdk-13_linux-x64_bin.tar.gz
paul@paul:~/tmp$ tar zxfv openjdk-13_linux-x64_bin.tar.gz
paul@paul:~/tmp$ ./jdk-13/bin/java --version
**openjdk 13 2019-09-17**
OpenJDK Runtime Environment (build 13+33)
OpenJDK 64-Bit Server VM (build 13+33, mixed mode, sharing)

Compared to 11 version:
paul@paul:~/tmp$ java --version
**openjdk 11.0.3 2019-04-16**
OpenJDK Runtime Environment (build 11.0.3+7-Ubuntu-1ubuntu218.04.1)
OpenJDK 64-Bit Server VM (build 11.0.3+7-Ubuntu-1ubuntu218.04.1, mixed mode, sharing)


It seems that there is no more 13.x.x convention in naming.
So what is Your advice
should it be simply 13 or 13.0.0 ?

marc0der:
Yep, let's stick to the original versioning scheme as outlined by [JEP-223](http://openjdk.java.net/jeps/223), so 13.0.0 it is for this GA release. Any subsequent security release will be 13.0.1 etc.

javafx-dev:
Ok but in this JEP-223 we can read:

...
**'The version number does not include trailing zero elements; i.e., $SECURITY is omitted if it has the value zero, and $MINOR is omitted if both $MINOR and $SECURITY have the value zero.'**
...
So if we want to stick to this scheme we it should be 13 not 13.0.0

What do You think?

marc0der:
13.0.0

javafx-dev:
Done.

PR #157, opened 2019-09-19
Add vendor - OpenJDK

javafx-dev:


marc0der:
Database migrations are immutable, once they have run they can't be changed.

Usually, you would need to write a migration that deletes and then reinserts the offending entry. This time, I've made a manual db update, but please keep this in mind for the future.

javafx-dev:
You did manual update but do You think that local installation will have proper vendor set or it will be unknown? maybe its better to merge this pull request?

PR #161, opened 2019-10-03
Update AdoptOpenJDK candidates

vpavic:
This PR updates AdoptOpenJDK candidates:
- updates `12.0` to current versions
- adds `13.0`

### DIFF
--- a/src/main/scala/io/sdkman/changelogs/JavaMigrations.scala
+++ b/src/main/scala/io/sdkman/changelogs/JavaMigrations.scala
@@ -659,4 +659,57 @@ class JavaMigrations {
       .insert()
     Seq(Linux64, MacOSX, Windows).foreach(removeVersion("java", "14.ea.14-open", _))
   }
+
+  @ChangeSet(order = "190", id = "190-add_12.0.2.hs-adpt", author = "vpavic")
+  def migrate190(implicit db: MongoDatabase): Unit = {
+    val candidate = "java"
+    val oldVersion = "12.0.1.hs-adpt"
+    val newVersion = "12.0.2.hs-adpt"
+    Seq(Linux64, MacOSX, Windows).foreach(removeVersion(candidate, oldVersion, _))
+    List(
+      Version(candidate, newVersion, "https://github.com/AdoptOpenJDK/openjdk12-binaries/releases/download/jdk-12.0.2%2B10/OpenJDK12U-jdk_x64_linux_hotspot_12.0.2_10.tar.gz", Linux64, Some(AdoptOpenJDK)),
+      Version(candidate, newVersion, "https://github.com/AdoptOpenJDK/openjdk12-binaries/releases/download/jdk-12.0.2%2B10.3/OpenJDK12U-jdk_x64_mac_hotspot_12.0.2_10.tar.gz", MacOSX, Some(AdoptOpenJDK)),
+      Version(candidate, newVersion, "https://github.com/AdoptOpenJDK/openjdk12-binaries/releases/download/jdk-12.0.2%2B10/OpenJDK12U-jdk_x64_windows_hotspot_12.0.2_10.zip", Windows, Some(AdoptOpenJDK)))
+      .validate()
+      .insert()
+  }
+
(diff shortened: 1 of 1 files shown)


src/main/scala/io/sdkman/changelogs/JavaMigrations.scala
```diff
@@ -659,4 +659,39 @@ class JavaMigrations {
...
     Seq(Linux64, MacOSX, Windows).foreach(removeVersion("java", "14.ea.14-open", _))
   }
+
+  @ChangeSet(order = "190", id = "190-add_12.0.2.hs-adpt", author = "vpavic")
+  def migrate190(implicit db: MongoDatabase): Unit = {
+    Seq(Linux64, MacOSX, Windows).foreach(removeVersion("java", "12.0.1.hs-adpt", _))
+    Map(Linux64 -> "https://github.com/AdoptOpenJDK/openjdk12-binaries/releases/download/jdk-12.0.2%2B10/OpenJDK12U-jdk_x64_linux_hotspot_12.0.2_10.tar.gz",
+      MacOSX -> "https://github.com/AdoptOpenJDK/openjdk12-binaries/releases/download/jdk-12.0.2%2B10.3/OpenJDK12U-jdk_x64_mac_hotspot_12.0.2_10.tar.gz",
+      Windows -> "https://github.com/AdoptOpenJDK/openjdk12-binaries/releases/download/jdk-12.0.2%2B10/OpenJDK12U-jdk_x64_windows_hotspot_12.0.2_10.zip")
+      .foreach({ case (platform, url) => Version("java", "12.0.2.hs-adpt", url, platform, Some(AdoptOpenJDK)).validate().insert() })
```
marc0der: Mind if we use the type classes provided to do the iteration as per normal? The underlying type classes automatically iterate over `List[Version]` and provide all the necessary plumbing to do validation and insertion.

vpavic: Sure thing @marc0der, if you prefer that way.

I went this route mostly to reduce the duplication as candidate version string (e.g. `12.0.2.hs-adpt`) is declared only once per migration.

marc0der: Could you perhaps assign version to a val to make it safer?

vpavic: Yeah, that was also my plan when I get to this later today.

vpavic: The PR is now updated @marc0der. I was wondering could we have specializations of `Version`, so that we don't need to repeat candidate name each and every time? Something like `JavaVersion`, `GradleVersion` etc. Or at least constants for candidate names.

src/main/scala/io/sdkman/changelogs/JavaMigrations.scala
```diff
@@ -659,4 +659,57 @@ class JavaMigrations {
...
   }
+
+  @ChangeSet(order = "190", id = "190-add_12.0.2.hs-adpt", author = "vpavic")
+  def migrate190(implicit db: MongoDatabase): Unit = {
+    val candidate = "java"
+    val oldVersion = "12.0.1.hs-adpt"
+    val newVersion = "12.0.2.hs-adpt"
+    Seq(Linux64, MacOSX, Windows).foreach(removeVersion(candidate, oldVersion, _))
+    List(
+      Version(candidate, newVersion, "https://github.com/AdoptOpenJDK/openjdk12-binaries/releases/download/jdk-12.0.2%2B10/OpenJDK12U-jdk_x64_linux_hotspot_12.0.2_10.tar.gz", Linux64, Some(OpenJDK)),
```
marc0der: The vendor should be `AdoptOpenJDK`, as `OpenJDK` refers to binaries found on java.net.

vpavic: Argh, that's a _regression_ due to c&p-ing back to `List` approach. Should be all good now.

marc0der:
Thanks again for taking the time. I always appreciate your PRs very much.

PR #171, opened 2019-10-28
Update AdoptOpenJDK 8 & 11, add upstream OpenJDK 8 & 11 (Linux & Windows)

andrebrait:
**Note:** Unfortunately, AdoptOpenJDK doesn't build upstream binaries for macOS

### DIFF
--- a/src/main/scala/io/sdkman/changelogs/JavaMigrations.scala
+++ b/src/main/scala/io/sdkman/changelogs/JavaMigrations.scala
@@ -679,4 +679,68 @@ class JavaMigrations {
     ).validate().insert()
     Seq(Linux64, MacOSX, Windows).foreach(platform => removeVersion(candidate = "java", version = "11.0.4-amzn", platform))
   }
+
+  @ChangeSet(order = "205", id = "205-add_adoptopenjdk-hs_8_0_232", author = "andrebrait")
+  def migrate205(implicit db: MongoDatabase) = {
+    List(
+      Version("java", "8.0.232.hs-adpt", "https://github.com/AdoptOpenJDK/openjdk8-binaries/releases/download/jdk8u232-b09/OpenJDK8U-jdk_x64_linux_hotspot_8u232b09.tar.gz", Linux64, Some(AdoptOpenJDK)),
+      Version("java", "8.0.232.hs-adpt", "https://github.com/AdoptOpenJDK/openjdk8-binaries/releases/download/jdk8u232-b09/OpenJDK8U-jdk_x64_mac_hotspot_8u232b09.tar.gz", MacOSX, Some(AdoptOpenJDK)),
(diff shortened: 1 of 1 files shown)


src/main/scala/io/sdkman/changelogs/JavaMigrations.scala
```diff
@@ -679,4 +679,68 @@ class JavaMigrations {
...
+      Version("java", "11.0.5.j9-adpt", "https://github.com/AdoptOpenJDK/openjdk11-binaries/releases/download/jdk-11.0.5+10_openj9-0.17.0/OpenJDK11U-jdk_x64_windows_openj9_11.0.5_10_openj9-0.17.0.zip", Windows, Some(AdoptOpenJDK)))
+      .validate()
+      .insert()
+    Seq(Linux64, MacOSX, Windows).foreach(removeVersion("java", "11.0.4.j9-adpt", _))
+  }
+
+  @ChangeSet(order = "209", id = "209-add_openjdk_8_0_232", author = "andrebrait")
+  def migrate209(implicit db: MongoDatabase) = {
+    List(
+      Version("java", "8.0.232-open", "https://github.com/AdoptOpenJDK/openjdk8-upstream-binaries/releases/download/jdk8u232-b09/OpenJDK8U-jdk_x64_linux_8u232b09.tar.gz", Linux64),
```
marc0der: We can't really add adoptopenjdk builds as `-open` because this is reserved for reference implementation JDKs coming from http://jdk.java.net.

They aren't releasing any new versions of JDK8, and only seem to have new RIs for 11, 12 and 13 now.

src/main/scala/io/sdkman/changelogs/JavaMigrations.scala
```diff
@@ -679,4 +679,68 @@ class JavaMigrations {
...
+      Version("java", "8.0.232-open", "https://github.com/AdoptOpenJDK/openjdk8-upstream-binaries/releases/download/jdk8u232-b09/OpenJDK8U-jdk_x64_windows_8u232b09.zip", Windows))
+      .validate()
+      .insert()
+  }
+
+  @ChangeSet(order = "210", id = "210-add_openjdk_11_0_5", author = "andrebrait")
+  def migrate210(implicit db: MongoDatabase) = {
```
marc0der: As above

src/main/scala/io/sdkman/changelogs/JavaMigrations.scala
andrebrait: @marc0der Although the URLs are from AdoptOpenJDK, these are actually binaries from upstream builds of the OpenJDK (see https://adoptopenjdk.net/upstream.html and https://github.com/AdoptOpenJDK/TSC/issues/65)

So these are provided by AdoptOpenJDK on behalf of RedHat as builds of the reference implementation. We can find another name for them, if you want (perhaps `redhat`?) but these are technically the continuation of what is provided in http://jdk.java.net

marc0der: @andrebrait Thanks for the explanation, it makes perfect sense now. I think keeping it under the `-open` label works perfectly in that case.

Going forward, would it be an idea for the AdoptOpenJDK project to start using our [vendor API](https://sdkman.io/vendors) instead of using PRs? This could make things a lot easier than using a PR for each new release.

PR #174, opened 2019-10-29
Fix upstream OpenJDK not being listed as candidate for Java.net OpenJDK

andrebrait:


marc0der:
Thanks for fixing this!

PR #179, opened 2019-11-19
Update version of Zulu October release

dimitryc:


marc0der:
@dimitryc the `JavaMigrations` have now been deprecated, all new Java migrations going into one of the corresponding vendor-specific migrations files in the `java` package. Having split out to the new migration files, we now no longer need to delete old migrations.

marc0der:
@dimitryc Also, please read through the README file to understand how these migrations work and can be tested locally before submitting a new PR.

dimitryc:
Thanks for the comments. Moved migrations for versions of Zulu October releases to new migration file. Tested with local build and confirmed that these versions were added to "versions" collection.

marc0der:
All the migrations you moved from `JavaMigrations` already seem to be in the database. Perhaps not a good idea to move them into a new migration file as they will then be re-inserted.

Database migrations should be considered immutable, so we can't really move but only create new migrations. I think we should leave the old migrations in the `JavaMigrations` class, and subsequently add any new release migrations to the `AzulZuluMigrations` class. wdyt?

Do you currently have any new versions that you want to release?

jorsol:
The issue with this update is that this is a vendor patch update, so in general it's the same version eg.: `11.0.5-zulu` -> vendor version is `11.35.15`, and the version already inserted is `11.35.13`.

So unless there is a "patch" in the version like `11.0.5.1-zulu` it will conflict with whatever is already inserted, in other words the idea is to "update" the migration URL, but if the migrations are considered immutable it's going to be difficult to do this.

marc0der:
Thanks @jorsol, that makes perfect sense.What we could do is delete the existing version, then insert the new patched version. Unfortunately we can't accommodate your vendor specific patch version in our version scheme though (they all need to be consistent for various reasons).

The current migrations in this PR are trying to delete versions that are no longer there, then trying to insert over versions that are already there (in our version scheme). I suggest first deleting the existing version, _then_ inserting your new patched version (with the same version as the deleted one).

src/main/scala/io/sdkman/changelogs/java/AzulZuluMigrations.scala
```diff
@@ -2,8 +2,47 @@ package io.sdkman.changelogs.java
...
 @ChangeLog(order = "019")
 class AzulZuluMigrations {
 
+  @ChangeSet(order = "001", id = "001-add_zulu_7_0_242", author = "jorsol")
+  def migrate001(implicit db: MongoDatabase) = {
+    List(
+      Version("java", "7.0.242-zulu", "https://cdn.azul.com/zulu/bin/zulu7.34.0.5-ca-jdk7.0.242-linux_x64.tar.gz", Linux64, Some(Zulu)),
+      Version("java", "7.0.242-zulu", "https://cdn.azul.com/zulu/bin/zulu7.34.0.5-ca-jdk7.0.242-win_x64.zip", Windows, Some(Zulu))
+    ).validate().insert()
+    Seq(Linux64, Windows, MacOSX).foreach(platform => removeVersion(candidate = "java", version = "7.0.232-zulu", platform))
```
marc0der: This version is no longer in the database. If you are trying to perform deletion of the superseded `7.0.242-zulu` with a patched version, move this statement above the inserts and then try inserting your new versions after that.

src/main/scala/io/sdkman/changelogs/java/AzulZuluMigrations.scala
```diff
@@ -2,8 +2,47 @@ package io.sdkman.changelogs.java
...
+  }
+
+  @ChangeSet(order = "002", id = "002-add_zulu_8_0_232", author = "jorsol")
+  def migrate002(implicit db: MongoDatabase) = {
+    List(
+      Version("java", "8.0.232-zulu", "https://cdn.azul.com/zulu/bin/zulu8.42.0.23-ca-jdk8.0.232-linux_x64.tar.gz", Linux64, Some(Zulu)),
+      Version("java", "8.0.232-zulu", "https://cdn.azul.com/zulu/bin/zulu8.42.0.23-ca-jdk8.0.232-win_x64.zip", Windows, Some(Zulu)),
+      Version("java", "8.0.232-zulu", "https://cdn.azul.com/zulu/bin/zulu8.42.0.23-ca-jdk8.0.232-macosx_x64.tar.gz", MacOSX, Some(Zulu))
+    ).validate().insert()
+    Seq(Linux64, Windows, MacOSX).foreach(platform => removeVersion(candidate = "java", version = "8.0.222-zulu", platform))
```
marc0der: As above.

src/main/scala/io/sdkman/changelogs/java/AzulZuluMigrations.scala
```diff
@@ -2,8 +2,47 @@ package io.sdkman.changelogs.java
...
+  }
+
+  @ChangeSet(order = "003", id = "003-add_zulu_11_0_5", author = "jorsol")
+  def migrate003(implicit db: MongoDatabase) = {
+    List(
+      Version("java", "11.0.5-zulu", "https://cdn.azul.com/zulu/bin/zulu11.35.15-ca-jdk11.0.5-linux_x64.tar.gz", Linux64, Some(Zulu)),
+      Version("java", "11.0.5-zulu", "https://cdn.azul.com/zulu/bin/zulu11.35.15-ca-jdk11.0.5-win_x64.zip", Windows, Some(Zulu)),
+      Version("java", "11.0.5-zulu", "https://cdn.azul.com/zulu/bin/zulu11.35.15-ca-jdk11.0.5-macosx_x64.tar.gz", MacOSX, Some(Zulu))
+    ).validate().insert()
+    Seq(Linux64, Windows, MacOSX).foreach(platform => removeVersion(candidate = "java", version = "11.0.4-zulu", platform))
```
marc0der: As above.

src/main/scala/io/sdkman/changelogs/java/AzulZuluMigrations.scala
```diff
@@ -2,8 +2,47 @@ package io.sdkman.changelogs.java
...
+  }
+
+  @ChangeSet(order = "004", id = "004-add_zulu_13_0_1", author = "jorsol")
+  def migrate004(implicit db: MongoDatabase) = {
+    List(
+      Version("java", "13.0.1-zulu", "https://cdn.azul.com/zulu/bin/zulu13.28.11-ca-jdk13.0.1-linux_x64.tar.gz", Linux64, Some(Zulu)),
+      Version("java", "13.0.1-zulu", "https://cdn.azul.com/zulu/bin/zulu13.28.11-ca-jdk13.0.1-win_x64.zip", Windows, Some(Zulu)),
+      Version("java", "13.0.1-zulu", "https://cdn.azul.com/zulu/bin/zulu13.28.11-ca-jdk13.0.1-macosx_x64.tar.gz", MacOSX, Some(Zulu))
+    ).validate().insert()
+    Seq(Linux64, Windows, MacOSX).foreach(platform => removeVersion(candidate = "java", version = "13.0.0-zulu", platform))
```
marc0der: As above.

dimitryc:
Thanks for explaining. New patch reverts previous changes and creates new migrations. New migration first removes versions (8.0.232-zulu with vendor update 8.42.0.21) and then creates new versions (8.0.232-zulu with vendor update 8.42.0.23). Similar migration for 11.0.5-zulu.

Testing: switching to master branch and running migrations (db now contains versions with older vendor update). Switching to new branch and running migrations (db now contains newer versions).

src/main/scala/io/sdkman/changelogs/java/AzulZuluMigrations.scala
```diff
@@ -2,8 +2,28 @@ package io.sdkman.changelogs.java
 
 import com.github.mongobee.changeset.{ChangeLog, ChangeSet}
 import com.mongodb.client.MongoDatabase
+import io.sdkman.changelogs.{Linux64, MacOSX, Version, Windows, Zulu, removeVersion}
 
 @ChangeLog(order = "019")
```
jorsol: This order id has any meaning?

marc0der: Yep, the order that migrations are run in if they haven't been run before.

PR #180, opened 2019-11-20
Add GraalVM 19.3.0

eddumelendez:


src/main/scala/io/sdkman/changelogs/java/GraalVmMigrations.scala
```diff
@@ -2,8 +2,52 @@ package io.sdkman.changelogs.java
...
 
 @ChangeLog(order = "021")
 class GraalVmMigrations {
 
+	@ChangeSet(order = "001", id = "001-add_graalvm_19_3_0", author = "eddumelendez")
+	def migrate001(implicit db: MongoDatabase) = {
+		List(
+			Version(
+				candidate = "java",
+				version = "19.3.0-grl_java11",
```
marc0der: Unfortunately this version won't work. We'll need to find a better way of expressing this.

eddumelendez: Thanks for the feedback @marc0der. WDYT about having another attribute like `candidate_version` in `Version`? I think this will solve part of the problem. From the other side, it should be reflected into the Java list view.

eddumelendez: may something simple like version `19.3.0_java11-grl` can work? WDYT?

marc0der: Adding another attribute has huge impact on the domain, and will ripple through all the layers right down into the db. It can be done but is an expensive change. Not really worth doing for a single vendor of a single candidate (remember this affects all candidates on sdkman).

We are currently discussing some other [options here](https://github.com/sdkman/sdkman-db-migrations/issues/181). Let me know your thoughts.

rsvoboda: @eddumelendez `19.3.0.r11-grl` and `19.3.0.r8-grl` were agreed in the above linked issue

eddumelendez: thank for the update @rsvoboda!

eddumelendez:
@marc0der PR updated with he version agreed

marc0der:
Thank you @eddumelendez!

PR #187, opened 2019-12-04
add graalvm 19.3.0

lrlucena:
Add final version of graalvm 19.3.0.


marc0der:
Hi @lrlucena, we already have graalvm 19.3.0 final release for both Java 8 and 11. They have the labels of
* 19.3.0.r8-grl
* 19.3.0.r11-grl
respectively, as we don't really have another way of representing this.
Will close this issue for now. Please shout if this doesn't work for you.

PR #198, opened 2020-01-16
Added java 7 update 252 and zulu 14.0.0

philiplourandos:


philiplourandos:
@marc0der I've rebased with the latest from master. Added zulu JDK 14 as well

PR #208, opened 2020-01-23
Added AdoptOpenJDK 8.0.242(OpenJ9)

poad:


src/main/scala/io/sdkman/changelogs/java/AdoptOpenJdkMigrations.scala
```diff
@@ -178,4 +178,39 @@ class AdoptOpenJdkMigrations {
...
+      Version(
+        "java",
+        "8.0.242.j9-adpt",
+        "https://github.com/AdoptOpenJDK/openjdk8-binaries/releases/download/jdk8u242-b08_openj9-0.18.1/OpenJDK8U-jdk_x64_windows_openj9_8u242b08_openj9-0.18.1.zip",
+        Windows,
+        Some(AdoptOpenJDK)
+      )
+    ).validate()
+      .insert()
+    setCandidateDefault("java", "8.0.242.j9-adpt")
```
marc0der: Java 8 is not the default on sdkman. We track AdoptOpenJDK 11 HotSpot as our default candidate. Currently it is `11.0.6.hs-adpt` as per migration 0005 above.

poad: OK!

PR #216, opened 2020-02-06
Add karaf to be installable via sdkman.

tonit:
Apache Karaf is a polymorphic application runtime (https://karaf.apache.org/) and is used in a similar area as the already existing "container frameworks" springboot, micronaut etc.

### DIFF
--- a/src/main/scala/io/sdkman/changelogs/KarafMigrations.scala
+++ b/src/main/scala/io/sdkman/changelogs/KarafMigrations.scala
@@ -1,0 +1,29 @@
+package io.sdkman.changelogs
+
+import com.github.mongobee.changeset.{ChangeLog, ChangeSet}
+import com.mongodb.client.MongoDatabase
+
+@ChangeLog(order = "025")
+class KarafMigrations {
+
+  @ChangeSet(order = "001", id = "001_add_karaf_candidate", author = "tonit")
+  def migration001(implicit db: MongoDatabase) = {
+    Candidate(
+      candidate = "karaf",
+      name = "Apache Karaf",
+      description = "Karaf is a polymorphic applications-runtime.",
+      websiteUrl = "https://karaf.apache.org/",
+      distribution = "UNIVERSAL"
+    ).insert()
+  }
+
+  @ChangeSet(order = "002", id = "002-add_karaf_4.2.8", author = "tonit")
+  def migration002(implicit db: MongoDatabase) = {
+    Version(
+      "karaf",
+      "4.2.8",
+      "https://archive.apache.org/dist/karaf/4.2.8/apache-karaf-4.2.8.tar.gz"
+    ).insert()
+      .asCandidateDefault()
+  }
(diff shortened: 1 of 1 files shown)


tonit:
@marc0der I have actually no idea how i can test this myself since the docs only speak about the migration itself (which works). How would I do
> sdk install karaf 
against a local backend? 
Thanks,
Toni

tonit:
I am missing clues on how I tell sdkman where the binaries are within the zip file..

marc0der:
Hi @tonit, this PR is only adding a new version but not the actual candidate itself. You can refer to [this section](https://github.com/sdkman/sdkman-db-migrations#adding-a-new-candidate) of the README to find out how to add the candidate itself.

Usually, the zip will have a structure as follows:

```
karaf-4.2.8
        /bin
                 /mybinary.bat
                 /mybinary
        /libexec
        /etc
```
So your zip would have a containing folder that in turn has a bin folder with all binaries. You could look at any other sdk like the JDK, Groovy, Scala, Spark for examples if you need.

Hope that helps!

tonit:
I have added the Candidate change and amended my first change. 
About the bin-structure, that looks fine as it exactly fits the current structure.

tonit:
Hey @marc0der how is the process going now? Is there anything I can do?

marc0der:
Hi @tonit, apologies but I've been a bit busy of late. I'll do a final review of your PR over the weekend.

src/main/scala/io/sdkman/changelogs/KarafMigrations.scala
```diff
@@ -0,0 +1,29 @@
...
+
+@ChangeLog(order = "025")
+class KarafMigrations {
+
+  @ChangeSet(order = "001", id = "001_add_karaf_candidate", author = "tonit")
+  def migration001(implicit db: MongoDatabase) = {
+    Candidate(
+      candidate = "karaf",
+      name = "Apache Karaf",
+      description = "Karaf is a polymorphic applications-runtime.",
```
marc0der: Would you please be able to provide something slightly more descriptive for this field? You can see examples of what the other projects provide when you issue the `sdk list` command.

src/main/scala/io/sdkman/changelogs/KarafMigrations.scala
```diff
@@ -0,0 +1,29 @@
...
+import com.mongodb.client.MongoDatabase
+
+@ChangeLog(order = "025")
+class KarafMigrations {
+
+  @ChangeSet(order = "001", id = "001_add_karaf_candidate", author = "tonit")
+  def migration001(implicit db: MongoDatabase) = {
+    Candidate(
+      candidate = "karaf",
+      name = "Apache Karaf",
```
marc0der: Just `Karaf` will do.

tonit:
Thanks for reviewing this, @marc0der. I have applied the suggested changes.

PR #228, opened 2020-03-17
Retry adding 13.0.2-open

kishida:
There is no JDK13 in java.net repository now.
Due to the lack of `insert`, 13.0.2-open didn't be inserted.
https://github.com/sdkman/sdkman-db-migrations/blob/master/src/main/scala/io/sdkman/changelogs/java/OpenJdkMigrations.scala#L523

Should I remove the ineffectual code in `018-add_openjdk_java_13.0.2`?
https://github.com/sdkman/sdkman-db-migrations/blob/master/src/main/scala/io/sdkman/changelogs/java/OpenJdkMigrations.scala#L525-L547

### DIFF
--- a/src/main/scala/io/sdkman/changelogs/java/OpenJdkMigrations.scala
+++ b/src/main/scala/io/sdkman/changelogs/java/OpenJdkMigrations.scala
@@ -1001,4 +1001,35 @@ class OpenJdkMigrations {
       removeVersion("java", "14.ea.36-open", _)
     )
   }
+  @ChangeSet(
+    order = "033",
+    id = "033-retry-add_openjdk_java_13.0.2",
+    author = "poad, kishida"
+  )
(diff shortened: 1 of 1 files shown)


kishida:
Confirm with the code below.
`13.0.2-open` appeared. (and `13.ea.33-open`)

```
import com.mongodb.client.MongoClients;
import static com.mongodb.client.model.Filters.*;

public class ConnectSample {
    public static void main(String[] args) {
        var client = MongoClients.create();
        var db = client.getDatabase("sdkman");
        var coll = db.getCollection("versions");
        var ite = coll.find(and(
                eq("candidate", "java"), 
                regex("version", "13.+-open")));
        
        for (var doc : ite) {
            System.out.println(doc);
        }

        client.close();
    }
}
```

marc0der:
Thanks for getting to the bottom of that. Much appreciated!

PR #234, opened 2020-03-24
feat: add dotty candidate

mingchuno:

PR #236, opened 2020-03-27
fix(dotty): remove missing dotty 0.21.0 release

mingchuno:
fix #234 
cc @marc0der


marc0der:
Just installed it and it is broken on linux :-(

Looks like you referenced tarballs, not zips...

marc0der:
@mingchuno I've removed them from the database manually for now.

PR #248, opened 2020-04-19
Add muleflowdiagrams candidate with first version

manikmagar:


src/main/scala/io/sdkman/changelogs/MuleFlowDiagramsMigrations.scala
```diff
@@ -0,0 +1,38 @@
...
+      distribution = "UNIVERSAL"
+    ).insert()
+  }
+
+  @ChangeSet(
+    order = "002",
+    id = "002_add_muleflowdiagrams_0_20_0_1",
+    author = "manikmagar"
+  )
+  def migration002(implicit db: MongoDatabase) = {
```
marc0der, line 30: I suggest we leave out this migration as this will be done using the API shortly. Let's first just get the candidate added.

manikmagar: Hi @marc0der , I have updated PR to remove this migration.

src/main/scala/io/sdkman/changelogs/MuleFlowDiagramsMigrations.scala
```diff
@@ -0,0 +1,24 @@
...
+class MuleFlowDiagramsMigrations {
+
+  @ChangeSet(
+    order = "001",
+    id = "001_add_muleflowdiagrams_candidate",
+    author = "manikmagar"
+  )
+  def migration001(implicit db: MongoDatabase): Candidate = {
+    Candidate(
+      candidate = "muleflowdiagrams",
```
marc0der, line 16: I'm concerned that this is too verbose for our users. Could `mfd` or `mulefd` be better? This will be used as the candidate identifier that our users use on the CLI, the name will still be the same in all other contexts, listings etc.

src/main/scala/io/sdkman/changelogs/MuleFlowDiagramsMigrations.scala
```diff
@@ -0,0 +1,24 @@
+package io.sdkman.changelogs
+
+import com.github.mongobee.changeset.{ChangeLog, ChangeSet}
+import com.mongodb.client.MongoDatabase
+
+@ChangeLog(order = "027")
```
marc0der, line 6: Can you bump this to `028` please?

src/main/scala/io/sdkman/changelogs/MuleFlowDiagramsMigrations.scala
manikmagar: Sure, sounds good. I will change it to `mulefd`.

PR #255, opened 2020-04-27
Update OpenJDK 8 and 11 to latest upstream versions

andrebrait:


andrebrait:
@marc0der unfortunately, they only provide builds for Windows and Linux

andrebrait:
@marc0der see https://adoptopenjdk.net/upstream.html

marc0der:
Thanks!

PR #256, opened 2020-04-27
Fix the SapMachine JDK for macOS download URL was wrong

poad:
The download URL for '11.0.7-sapmchn' for macOS was from JRE.
Delete it and add the URL of the JDK.


tinobino:
Thanks guys! It is working now!

PR #261, opened 2020-05-19
Add GraalVM 20.1.0

eddumelendez:


marc0der:
Are there any older versions that we can drop here that aren't being used by Micronaut @ilopmar?

ilopmar:
This is how it looks like now:

```
 GraalVM       |     | 20.0.0.r11   | grl     | installed  | 20.0.0.r11-grl      
               |     | 20.0.0.r8    | grl     | installed  | 20.0.0.r8-grl       
               |     | 19.3.1.r11   | grl     | installed  | 19.3.1.r11-grl      
               |     | 19.3.1.r8    | grl     | installed  | 19.3.1.r8-grl       
               |     | 19.3.0.r11   | grl     |            | 19.3.0.r11-grl      
               |     | 19.3.0.r8    | grl     |            | 19.3.0.r8-grl       
               |     | 19.3.0.2.r11 | grl     |            | 19.3.0.2.r11-grl    
               |     | 19.3.0.2.r8  | grl     |            | 19.3.0.2.r8-grl     
               |     | 19.2.1       | grl     |            | 19.2.1-grl          
               |     | 19.1.1       | grl     |            | 19.1.1-grl          
               |     | 19.0.2       | grl     |            | 19.0.2-grl          
               |     | 1.0.0        | grl     |            | 1.0.0-rc-16-grl 
```

Micronaut supports officially `20.0.0` but I'm pretty sure it works with `19.3.1` and probably some older versions.

I think at least we can get rid of:
```
1.0.0-rc-16-grl 
19.0.2-grl        
19.1.1-grl        
19.2.1-grl  
19.3.0.r8-grl       
19.3.0.r11-grl     
``` 

And we could also probably remove:
```
19.3.0.2.r8-grl 
19.3.0.2.r11-grl    
```               

GraalVM improves a lot in every release and there is no point to support old releases. If folks want to use it, they should move to newer versions.

eddumelendez:
I have added a new commit in order to remove the older versions. Just let me know If there should be another change.

marc0der:
Thanks guys!

PR #266, opened 2020-06-04
Liberica mission control

den1ska:
Adding a new candidate `jmc`: https://openjdk.java.net/projects/jmc and its custom build LMC by BellSoft.


marc0der:
Thanks for the PR. Just a few comments:

- could you also create the candidate in a separate commit please? The current PR only adds versions but doesn't set up the candidate.
- would you be able to add versions for the java.net RI too please? It is just fair to our users not to be tied in to one specific vendor's versions.

den1ska:
@marc0der, Thank you for comments! I've just created a new PR for adding JMC RI and I'll update the current PR soon.

den1ska:
We decided to postpone adding LMC into SKDMAN until 7.1.2 is released.

PR #269, opened 2020-06-15
Added JMC 7.0.1 General-Availability Release.

den1ska:
Added JDK Mission Control 7.

Please note that: JDK Mission Control is currently in development. Early-Access binaries are available from Oracle. I'm not sure if I correctly chose the vendor name.

### DIFF
--- a/src/main/scala/io/sdkman/changelogs/jmc/OpenJDKJmcMigrations.scala
+++ b/src/main/scala/io/sdkman/changelogs/jmc/OpenJDKJmcMigrations.scala
@@ -1,0 +1,50 @@
+package io.sdkman.changelogs.jmc
+
+import com.github.mongobee.changeset.{ChangeLog, ChangeSet}
+import com.mongodb.client.MongoDatabase
+import io.sdkman.changelogs.{Linux64, MacOSX, Version, Windows, _}
+
+@ChangeLog(order = "029")
+class OpenJDKJmcMigrations {
+
+  @ChangeSet(order = "001", id = "001_add_jmc_candidate", author = "den1ska")
+  def migration001(implicit db: MongoDatabase) = {
+    Candidate(
+      candidate = "jmc",
+      name = "JDK Mission Control",
+      description =
+        "JDK Mission Control 7 (JMC) is an advanced set of tools that enables efficient and detailed data analysis, and delivers advanced, unobtrusive Java monitoring and management. JMC provides sections for common analysis areas such as code performance, memory, and latency.",
+      websiteUrl = "https://jdk.java.net/jmc/",
+      distribution = "UNIVERSAL"
+    ).insert()
+  }
+
+  @ChangeSet(order = "002", id = "002-add_jmc_7.0.1-ga", author = "den1ska")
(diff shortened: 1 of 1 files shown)


src/main/scala/io/sdkman/changelogs/jmc/OpenJDKJmcMigrations.scala
```diff
@@ -0,0 +1,50 @@
...
+
+  @ChangeSet(order = "001", id = "001_add_jmc_candidate", author = "den1ska")
+  def migration001(implicit db: MongoDatabase) = {
+    Candidate(
+      candidate = "jmc",
+      name = "JDK Mission Control",
+      description =
+        "JDK Mission Control 7 (JMC) is an advanced set of tools that enables efficient and detailed data analysis, and delivers advanced, unobtrusive Java monitoring and management. JMC provides sections for common analysis areas such as code performance, memory, and latency.",
+      websiteUrl = "https://jdk.java.net/jmc/",
+      distribution = "UNIVERSAL"
```
marc0der, line 18: This looks more like `PLATFORM_SPECIFIC` to me. wdyt?

src/main/scala/io/sdkman/changelogs/jmc/OpenJDKJmcMigrations.scala
```diff
@@ -0,0 +1,50 @@
...
+import com.mongodb.client.MongoDatabase
+import io.sdkman.changelogs.{Linux64, MacOSX, Version, Windows, _}
+
+@ChangeLog(order = "029")
+class OpenJDKJmcMigrations {
+
+  @ChangeSet(order = "001", id = "001_add_jmc_candidate", author = "den1ska")
+  def migration001(implicit db: MongoDatabase) = {
+    Candidate(
+      candidate = "jmc",
```
marc0der, line 13: That name looks fine for me :+1:

alshamiri5 (approved):

marc0der (changes requested):
This requires some work in the backend. Either we do the real work to seamlessly handle tarballs for any candidate, or we do a dirty hack to correct this case in sdkman-hooks. I prefer the former, but have limited availability to do the work right now. Any takers?

vektory79:
Just note there, that alternative disrtibutions of the JMC is present:

* https://bell-sw.com/pages/lmc-7.1.1/
* https://jp.azul.com/products/zulu-mission-control/

So versioning should take this into account.

iemejia:
+ https://adoptopenjdk.net/jmc.html

src/main/scala/io/sdkman/changelogs/jmc/OpenJDKJmcMigrations.scala
reinhapa: @marc0der I would think too. Also there are multiple vendors involved as in the Java JDKs...

marc0der: This is a very old PR. JMC is long since available on SDKMAN, all though nobody seems to be using it.

Issue #193, opened 2020-01-06
Add parameters to Spark candidates


hammer:
Some pre-built Spark releases are built with support for different versions of Scala and/or Hadoop.

For example, there are 4 different [Spark 2.4.4 pre-built releases](https://archive.apache.org/dist/spark/spark-2.4.4/):

| File name                                     | Scala | Hadoop |
|-----------------------------------------------|-------|--------|
| spark-2.4.4-bin-without-hadoop.tgz            | 2.11  | None   |
| spark-2.4.4-bin-hadoop2.6.tgz                 | 2.11  | 2.6    |
| spark-2.4.4-bin-hadoop2.7.tgz                 | 2.11  | 2.7+   |
| spark-2.4.4-bin-without-hadoop-scala-2.12.tgz | 2.12  | None   |

It would be nice if sdkman knew about these options.

marc0der:
This could be handled in a simple way by using the version string. The max length is 15 so we could try using that. In other words something like `2.4.4_2.11_2.6`.

Sdkman is completely agnostic of the candidates it serves (except in the case of Java), and I really don't want to hack the core for candidate specific exceptions.

PR #274, opened 2020-06-22
Add Apache Spark 2.4.6

flvndh:


flvndh:
I just saw this issue #193 after submitting the PR. Should we follow this nomenclature from now on ?

marc0der:
It wasn't really followed through with a PR, so let's keep it as it is for now.

PR #276, opened 2020-06-28
Add AdoptOpenJDK 8, 11 (HotSpot) & Corretto 8, 11 for Linux ARM64

dvdkruk:
Add AdoptOpenJDK 8 & 11 (HotSpot) and Corretto 8 & 11 (thanks to @pnavais) as the first four candidates for Linux ARM64. Others one can be added later.

Part of: sdkman/sdkman-cli#769


src/main/scala/io/sdkman/changelogs/package.scala
```diff
@@ -154,6 +154,14 @@ package object changelogs {
     override val id = "LINUX_32"
   }
 
+  case object LinuxARM extends Platform {
```
marc0der: I'd suggest `LinuxARM32` to make it more explicit.

src/main/scala/io/sdkman/changelogs/package.scala
```diff
@@ -154,6 +154,14 @@ package object changelogs {
...
+    override val id = "LINUX_ARM"
```
marc0der: And how about `LINUX_ARM_32`?

src/main/scala/io/sdkman/changelogs/package.scala
```diff
@@ -154,6 +154,14 @@ package object changelogs {
...
+  }
+
+  case object LinuxARM64 extends Platform {
+    override val id = "LINUX_ARM64"
```
marc0der, line 158: Same as above.

src/main/scala/io/sdkman/changelogs/java/AdoptOpenJdkMigrations.scala
```diff
@@ -564,4 +564,28 @@ class AdoptOpenJdkMigrations {
   def migrate0017(implicit db: MongoDatabase) =
     setCandidateDefault("java", "11.0.7.hs-adpt")
 
+  @ChangeSet(
+    order = "0018",
+    id = "0018-add_adoptopenjdk-8-11-linuxarm64",
+    author = "dvdkruk"
+  )
+  def migrate0018(implicit db: MongoDatabase) = {
```
marc0der, line 572: We would need to bump this changeset to `0019`.

marc0der:
Apart from bumping the migration, LGTM.

marc0der:
@dvdkruk It is probably best to leave this PR until last, as we don't want to publish anything that we aren't equipped to serve yet.

marc0der:
Look like master got merged into this branch and is causing all interim commits to show up too :point_up:

Could you try a rebase instead of merge please?

dvdkruk:
@marc0der you are right fixed it now and also upped the migration version.

marc0der:
@dvdkruk I extracted your changes as a diff, then applied that directly to master as a new commit.

I think the main problem occurred when master was merged back into this feature branch. As a rule, I generally always rebase branches on master. Conversely I merge branches back into master. This approach generally seems to work best for me.

Anyway, all is good now so we can probably close this PR. Would you be willing to test if you can download and install an ARM64 JDK as I don't have any such hardware available?

dvdkruk:
Merged in bdf3446bd77a4aeb2be3c725b0825224dc8691c1

Testing the build now and it looks good so far, will test the other version now: 
image

PR #284, opened 2020-07-09
openjdk 15ea31 and 16ea5

eddumelendez:
- Add OpenJDK 15ea31
- Add OpenJDK 16ea5


eddumelendez:
@marc0der broke the build at #287 😢 I have updated this PR with the fix

marc0der:
I fixed it (and also made some optimisations). Please let me know how you like the change.
Shall we do subsequent migrations in this style going forward? wdyt?

eddumelendez:
thanks for fixing it! and the change looks much better! I will update my PR

eddumelendez:
@marc0der PR updated. hopefully no syntax error this time 🙈

marc0der:
Looks great! I suggest we just keep reusing these three constructs from now on (in other words, increment the `ChangeSet`s and reuse these three blocks of code without accumulation). In the end, version control will give us all the history we require should we need it. How does it sound?

PR #286, opened 2020-07-14
Add recent Apache Ant versions

andrebrait:
I have no idea how this is going to look like with the existing versions of ant in the database, though. They're not in this repository. I'm mimicking what was doen with Maven here (which also has versions that are not here in the migrations)

### DIFF
--- a/src/main/scala/io/sdkman/changelogs/AntMigration.scala
+++ b/src/main/scala/io/sdkman/changelogs/AntMigration.scala
@@ -1,0 +1,203 @@
+package io.sdkman.changelogs
+
+import com.github.mongobee.changeset.{ChangeLog, ChangeSet}
+import com.mongodb.client.MongoDatabase
+
+@ChangeLog(order = "030")
+class AntMigration {
+  @ChangeSet(
+    order = "001",
+    id = "001-add_ant_1.9.10",
+    author = "andrebrait"
+  )
+  def migration001(implicit db: MongoDatabase): Unit = {
+    Version(
+      "ant",
+      "1.9.10",
+      "https://archive.apache.org/dist/ant/binaries/apache-ant-1.9.10-bin.zip"
+    ).validate()
+      .insert()
+      .asCandidateDefault()
+  }
+
+  @ChangeSet(
+    order = "002",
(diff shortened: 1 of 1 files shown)


src/main/scala/io/sdkman/changelogs/AntMigration.scala
```diff
@@ -0,0 +1,203 @@
+package io.sdkman.changelogs
+
+import com.github.mongobee.changeset.{ChangeLog, ChangeSet}
+import com.mongodb.client.MongoDatabase
+
+@ChangeLog(order = "030")
```
marc0der, line 6: I think the next one in sequence is `032`

src/main/scala/io/sdkman/changelogs/AntMigration.scala
```diff
@@ -0,0 +1,203 @@
...
+      "1.9.14",
+      "https://archive.apache.org/dist/ant/binaries/apache-ant-1.9.14-bin.zip"
+    ).validate()
+      .insert()
+      .asCandidateDefault()
+  }
+
+  @ChangeSet(
+    order = "006",
+    id = "006-add_ant_1.9.14",
```
marc0der, line 85: Should that be `1.9.15`?

src/main/scala/io/sdkman/changelogs/AntMigration.scala
```diff
@@ -0,0 +1,203 @@
...
+    author = "andrebrait"
+  )
+  def migration001(implicit db: MongoDatabase): Unit = {
+    Version(
+      "ant",
+      "1.9.10",
+      "https://archive.apache.org/dist/ant/binaries/apache-ant-1.9.10-bin.zip"
+    ).validate()
+      .insert()
+      .asCandidateDefault()
```
marc0der, line 20: This only needs to be set once on the version that should be marked as default. How about only putting it on `1.10.8`?

marc0der:
Looks good apart from some minor tweaks!

andrebrait:
@marc0der I fixed the issues you pointed. Thanks for that!

marc0der:
Thank you!

PR #288, opened 2020-07-14
Add new versions of LibericaJDK

strizhik:


marc0der:
To prevent these ChangeLogs from growing out of control (and to keep the build times down), would it make sense to apply the same approach as what we have taken in #284 ?

Also, perhaps you could do a cleanup (removal) of old migrations that are no longer used.

Lastly, it would be preferable if you did a rebase against master rather than merging master into your feature branch, it keeps things a lot cleaner.

PR #292, opened 2020-07-16
Liberica 8u262, 11.0.8, 14.0.2

strizhik:
Hi @marc0der ,

I've addressed some of your comments from #288:
- implemented your suggestion from #284 
- got rid of merge

Can we do a housekeeping job a bit later? 
Should we move all necessary Liberica stuff from JavaMigrations to BellSoftLibericaMigrations?

### DIFF
--- a/src/main/scala/io/sdkman/changelogs/java/BellSoftLibericaMigrations.scala
+++ b/src/main/scala/io/sdkman/changelogs/java/BellSoftLibericaMigrations.scala
@@ -511,4 +511,208 @@ class BellSoftLibericaMigrations {
       removeVersion("java", "14.0.0.fx-librca", _)
     )
   }
+
+  @ChangeSet(
+    order = "0015",
+    id = "0015-add_bellsoft_8_0_262",
+    author = "strizhik"
+  )
+  def migrate0015(implicit db: MongoDatabase): Unit = {
+
+    Map(
+      Linux32 -> "bellsoft-jdk8u262+10-linux-i586.tar.gz",
+      Linux64 -> "bellsoft-jdk8u262+10-linux-amd64.tar.gz",
+      Windows -> "bellsoft-jdk8u262+10-windows-amd64.zip",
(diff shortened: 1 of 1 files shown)


marc0der:
I removed the obsolete migrations too. Thanks for contributing!

strizhik:
Hi @marc0der.
Thanks!

Could you please clarify one question? 
Since some of Liberica migrations located in JavaMigrations I think there is an issue with old migrations, located in JavaMigrations.

For example in current source code: 8.0.232-librca is added in https://github.com/sdkman/sdkman-db-migrations/blob/e77127940697845bd4cb447a337f257045a4f8b2/src/main/scala/io/sdkman/changelogs/JavaMigrations.scala#L1213 but not removed at the moment. Previously It was removed in https://github.com/sdkman/sdkman-db-migrations/commit/e589d88077604506f808e76fb0feb93065dd187b#diff-3ce664d6994ac74d8bca430d483841c9L41

Is this an issue from the sdkman point of view?

marc0der:
The moment a migration is merged to master, it will be run against the database once, and once only. After that it will be ignored on all subsequent runs. So after it has run, it is safe to remove it. Feel free to remove it from the `JavaMigrations` too if you feel up to it.

Hope I understood your question and that this explanation makes sense!

Cheers,
Marco.
GitHub <notifications@github.com> wrote:
“Hi @marc0der.
 Thanks!

Could you please clarify one question?
 Since some of Liberica migrations located in JavaMigrations I think there is an issue with old migrations, located in JavaMigrations.

For example in current source code: 8.0.232-librca is added in https://github.com/sdkman/sdkman-db-migrations/blob/e77127940697845bd4cb447a337f257045a4f8b2/src/main/scala/io/sdkman/changelogs/JavaMigrations.scala#L1213 but not removed at the moment. Previously It was removed in e589d88#diff-3ce664d6994ac74d8bca430d483841c9L41

Is this an issue from the sdkman point of view?”

PR #294, opened 2020-07-16
Added the SapMachine-11.0.8 and SapMachine-14.0.2

poad:


src/main/scala/io/sdkman/changelogs/java/SapMachineMigrations.scala
```diff
@@ -202,4 +202,56 @@ class SapMachineMigrations {
...
     ).validate().insert()
   }
+  @ChangeSet(
+    order = "0010",
+    id = "0010-add_sapmchn_jdk_11.0.8",
+    author = "poad"
+  )
+  def migrate0010(implicit db: MongoDatabase) =
+    Map(
+      Linux64 -> ("11.0.8", "linux-x64_bin.tar.gz"),
```
marc0der, line 212: The version doesn't change, so shall we just hard code it in the URL from now on? Then we can keep this as a simple map of KVs.

src/main/scala/io/sdkman/changelogs/java/SapMachineMigrations.scala
```diff
@@ -202,4 +202,56 @@ class SapMachineMigrations {
...
+        removeVersion("java", "11.0.7-sapmchn", version.platform)
+      }
+  @ChangeSet(
+    order = "0011",
+    id = "0011-add_sapmchn_jdk_14.0.2",
+    author = "poad"
+  )
+  def migrate0011(implicit db: MongoDatabase) =
+    Map(
+      Linux64 -> ("14.0.2", "linux-x64_bin.tar.gz"),
```
marc0der, line 238: Same as above.

marc0der: Otherwise it can just be declared as a `val` under the `case` statement.

src/main/scala/io/sdkman/changelogs/java/SapMachineMigrations.scala
poad: I see!

The version appears twice, so I made it logical, but it's certainly common as a URL, so it's more concise to hardcode it in one line.

src/main/scala/io/sdkman/changelogs/java/SapMachineMigrations.scala
poad: It is good Idea!
Change again.

poad:
@marc0der 
The points you pointed out have been corrected.

marc0der:
Sorry, jumped in there before you :smile: Already merged.

poad:
It's very educational.
He successfully applies what he thinks would be better to commonality.

marc0der:
Lol, nothing so lofty. IntelliJ just told me that it was seeing some duplicate code, so I factored it out into something reusable.

marc0der:
And Github seems confused too. It thinks that your branch hasn't been merged, even though it has. Will just close this PR then.

PR #297, opened 2020-07-18
Readd the OpenJDK 14.0.2

poad:
OpenJDK 14.0.2's build number has changed from 11 to 12.
The download URL has also been changed.

To change to the new download URL, remove the original changeset and add a new URL changeset.


marc0der:
I'm not entirely sure if that is an obsolete changeset. We have 14, 15 and 16 represented in this migration. The idea going forward is to simply update one of these with newer versions when a release of say 14.0.3 is made.

poad:
Sorry, I was wrong in my perception of the problem.

The build has been changed from 11 to 12 and the URL has changed.
This is why it fails to migrate.

I will review the revisions, including the title.

marc0der:
Thanks @poad!

PR #299, opened 2020-07-23
Added the ZuluFX 8.0.262/11.0.8/13.0.4/14.0.2

poad:


marc0der:
Thanks @poad. Is it worth removing the older obsolete versions per major version? That way we can keep the file size down.

marc0der:
Thanks @poad !

PR #303, opened 2020-07-30
Replace the Zulu 8u262 b17 to b19

poad:


marc0der:
Thanks! In future, we should probably reflect the extension like `b19` in the version. wdyt?

PR #308, opened 2020-07-31
Add LibericaJDK 8u265

strizhik:


src/main/scala/io/sdkman/changelogs/java/BellSoftLibericaMigrations.scala
```diff
@@ -8,23 +8,22 @@ import io.sdkman.changelogs.{Linux64, MacOSX, Version, Windows, _}
 class BellSoftLibericaMigrations {
 
   @ChangeSet(
-    order = "0015",
-    id = "0015-add_bellsoft_8_0_262",
+    order = "0016",
```
marc0der, line 11: ChangeSets are immutable like in most database migration tools. 16 was already used before.

I also suggest that for simplicity you keep updating versions like-for-like in the code below. In other words, simply bump JDK 8 to it's new version etc, and start a new number sequence accordingly. Hope it makes sense.

strizhik: Hi @marc0der,

I think Github's diff view a little bit confusing in this case.
I didn't change the ids of changesets. I've removed 15 and 19 from the middle of the file and added 21 and 22 to the end.

strizhik: Hi @marc0der,
Could you please merge this or #302 to get Lliberica 8u265 to be available in sdkman?

PR #311, opened 2020-08-07
Fix OpenJDK 15ea35

eddumelendez:
- Fix binaries for OpenJDK 15ea35
- Fix format


eddumelendez:
@marc0der sorry, this is the fix for the broken build 😢 thought I tested the links

PR #314, opened 2020-08-11
Update OpenJDK to 8.0.265 & 11.0.8

andrebrait:


src/main/scala/io/sdkman/changelogs/JavaMigrations.scala
```diff
@@ -508,6 +508,43 @@ class JavaMigrations {
     )
   }
 
+  @ChangeSet(order = "171", id = "171-add_bellsoft_12_0_2", author = "morgion")
```
marc0der: This class has been deprecated. New liberica releases should be in the dedicated changelog file.

andrebrait: This is not a new release and it's not part of the actual changes in this PR. I think you commented on the commit I made when reverting it?

I'll just squash and reopen the PR if that's easier.

marc0der:
Thank you!

PR #319, opened 2020-08-31
Adds Linux ARM64 librca versions 8,11,11fx,14,14fx

dvdkruk:
Adds LTS versions of Liberica for Linux ARM64 (AArch64)

PR #320, opened 2020-09-01
Correction of linux arm64 version labels of 11, 14

dvdkruk:
Hey @eddumelendez, found that I made a typo in #319 for the non FX versions. This results in alignment issues in the UI/CLI:
image
This PR removes the typos and replaces them with the right version ids.


marc0der:
Thanks, that has been merged now.

PR #321, opened 2020-09-01
Dev support for vscode on Linux ARM64

dvdkruk:
I had to do the following changes to make development with vscode on Linux ARM64 possible on this repo:
- Updated Scale to version 2.12.12 (min supported Scala version in metals)
- Updated README to use docker container `mongo:4.0` (as this version and up have ARM64 support)
- Git ignore `.vscode` dir

Contributing this back, so that the next person doesn't have to 😉

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,13 +9,13 @@ This enables the users of SDKMAN to contribute new Installation Candidates and r
 
 This repo uses [mongobee](https://github.com/mongobee/mongobee) as database migration framework. In order to contribute a PR, it is required to have a local installation of MongoDB on your machine. Alternatively run it up with Docker (works on Linux):
 
-        $ docker run -d --network=host --name=mongo mongo:3.2
+        $ docker run -d --network=host --name=mongo mongo:4.0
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -9,13 +9,13 @@ This enables the users of SDKMAN to contribute new Installation Candidates and r
 
 This repo uses [mongobee](https://github.com/mongobee/mongobee) as database migration framework. In order to contribute a PR, it is required to have a local installation of MongoDB on your machine. Alternatively run it up with Docker (works on Linux):
 
-        $ docker run -d --network=host --name=mongo mongo:3.2
+        $ docker run -d --network=host --name=mongo mongo:4.0
```
marc0der, line 12: The reason that we want to run it in 3.2 is to ensure that migrations are compatible with what we have in our prod environment. Mind if we leave this as it was?

marc0der:
Thanks!

README.md
dvdkruk: NP

PR #328, opened 2020-09-11
Add WindowsARM64 OpenJDK16ea10

eddumelendez:
link: https://twitter.com/JavaAtMicrosoft/status/1304204505165099008?s=20

@marc0der WDYT about this?

### DIFF
--- a/src/main/scala/io/sdkman/changelogs/java/OpenJdkMigrations.scala
+++ b/src/main/scala/io/sdkman/changelogs/java/OpenJdkMigrations.scala
@@ -9,6 +9,7 @@ import io.sdkman.changelogs.{
   OpenJDK,
   Version,
   Windows,
+  WindowsARM64,
   removeVersion
 }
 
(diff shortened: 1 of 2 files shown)


marc0der:
I think it's a super idea, but we will need to accommodate this on the backend first. Let's mark this PR as draft and I can look at getting the backend ready for this over the weekend.

marc0der:
I had a closer look at this over the weekend, and unfortunately this is a bit more complicated than anticipated. For starters, we need to get our hands on a system that is running arm64 architecture with windows. Then we need to run a shell on that box and figure out what the `uname` output is. This is passed to the backend and what we use to [select the platform](https://github.com/sdkman/sdkman-cli/blob/master/src/main/bash/sdkman-init.sh#L32) in the services. After that, we can go ahead and implement the new platform in the backend services.

lewurm:
> Then we need to run a shell on that box and figure out what the `uname` output is. This is passed to the backend and what we use to [select the platform](https://github.com/sdkman/sdkman-cli/blob/master/src/main/bash/sdkman-init.sh#L32) in the services.

Presumably you are using `uname` from a Cygwin install when dealing with Windows x86/x86_64, is that right? Cygwin is _not_ available for Windows ARM64 (except by running the x86 binaries via emulation, and I'm not sure if you want that).

WSL is available however, and `uname`/`uname -a` outputs this:
```console
$ uname
Linux
$ uname -a
Linux myhostname 4.19.104-microsoft-standard #1 SMP Wed Feb 19 06:45:06 UTC 2020 aarch64 aarch64 aarch64 GNU/Linux
```

marc0der:
So does that mean that the Linux arm64 binary would run perfectly without even needing the WindowsARM64 variant on WSL?

lewurm:
> So does that mean that the Linux arm64 binary would run perfectly without even needing the WindowsARM64 variant on WSL?

This is correct. Depending on your purposes this might be good enough. As soon as you need to cross the WSL boundary it can become tricky.

eddumelendez:
I think I can close this PR sin the Linux ARM will work as far as I understood. Thank you @lewurm for your help!

PR #343, opened 2020-09-20
Add OpenJDK 16ea6 Loom Early Access

soberich:
see https://docs.gradle.org/nightly/userguide/toolchains.html#sec:auto_detection

### DIFF
--- a/src/main/scala/io/sdkman/changelogs/java/OpenJdkMigrations.scala
+++ b/src/main/scala/io/sdkman/changelogs/java/OpenJdkMigrations.scala
@@ -190,7 +190,7 @@ class OpenJdkMigrations {
       }
 
   @ChangeSet(
-    order = "077",
+    order = "078",
     id = "077-add_openjdk_java_16-loom-6",
     author = "soberich"
   )
--- a/src/main/scala/io/sdkman/changelogs/java/OpenJdkMigrations.scala
+++ b/src/main/scala/io/sdkman/changelogs/java/OpenJdkMigrations.scala
@@ -212,7 +212,4 @@ class OpenJdkMigrations {
       .toList
       .validate()
       .insert()
-      .foreach { version =>
-        removeVersion("java", "16.0.6.lm-open", version.platform)
-      }
 }
--- a/src/main/scala/io/sdkman/changelogs/java/OpenJdkMigrations.scala
+++ b/src/main/scala/io/sdkman/changelogs/java/OpenJdkMigrations.scala
@@ -203,7 +203,7 @@ class OpenJdkMigrations {
         case (platform, binary) =>
           Version(
             "java",
-            "16.0.6.lm-open", //has to be "1.2.3" e.i. numeric in order to Gradle auto detection work https://docs.gradle.org/nightly/userguide/toolchains.html#sec:auto_detection 
+            "16.ea.6.lm-open",
             s"https://download.java.net/java/early_access/loom/6/$binary",
             platform,
             Some(OpenJDK)
--- a/src/main/scala/io/sdkman/changelogs/java/OpenJdkMigrations.scala
+++ b/src/main/scala/io/sdkman/changelogs/java/OpenJdkMigrations.scala
@@ -188,4 +188,33 @@ class OpenJdkMigrations {
       .foreach { version =>
         removeVersion("java", "16.ea.15-open", version.platform)
       }
+  
+  
+
+  @ChangeSet(
+    order = "077",
+    id = "077-add_openjdk_java_16-loom-6",
+    author = "soberich"
+  )
+  def migrate077(implicit db: MongoDatabase): Unit =
+    Map(
+      Linux64    -> "openjdk-16-loom+6-105_linux-x64_bin.tar.gz",
+      MacOSX     -> "openjdk-16-loom+6-105_osx-x64_bin.tar.gz",
+      Windows    -> "openjdk-16-loom+6-105_windows-x64_bin.zip"
+    ).map {
+        case (platform, binary) =>
+          Version(
+            "java",
+            "16.0.6-loom", //has to be "1.2.3" e.i. numeric in order to Gradle auto detection work https://docs.gradle.org/nightly/userguide/toolchains.html#sec:auto_detection 
+            s"https://download.java.net/java/early_access/loom/6/$binary",
+            platform,
+            Some(OpenJDK)
+          )
+      }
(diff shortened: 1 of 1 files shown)


src/main/scala/io/sdkman/changelogs/java/OpenJdkMigrations.scala
```diff
@@ -188,4 +188,31 @@ class OpenJdkMigrations {
...
+  def migrate077(implicit db: MongoDatabase): Unit =
+    Map(
+      Linux64    -> "openjdk-16-loom+6-105_linux-x64_bin.tar.gz",
+      MacOSX     -> "openjdk-16-loom+6-105_osx-x64_bin.tar.gz",
+      Windows    -> "openjdk-16-loom+6-105_windows-x64_bin.zip"
+    ).map {
+        case (platform, binary) =>
+          Version(
+            "java",
+            "16.0.6-loom", //has to be "1.2.3" e.i. numeric in order to Gradle auto detection work https://docs.gradle.org/nightly/userguide/toolchains.html#sec:auto_detection
```
marc0der, line 206: This suffix is not currently supported on our backend yet.

eddumelendez: what about `16.ea.6.loom-open`?

soberich: @eddumelendez "has to be `\d\.\d\.\d.+` at the moment.
@marc0der Could you please specify what exactly do you mean?

marc0der:
My main concern is that this is actually an ea build, and we are designating it as final release so that another tool like Gradle can detect it. I think it is better to make Gradle more flexible in its recognition.

soberich:
It is better, but until then this is a workaround to make this possible.

marc0der:
I'm happy for us to add loom, but then it should be called `16.ea.6.lm-open`. If that is a problem for Gradle, please do raise it with their team. They are a very responsive and friendly bunch.

eddumelendez:
I agree with @marc0der, we can add loom as any other openjdk candidate.

soberich:
> I'm happy for us to add loom, but then it should be called `16.ea.6.lm-open`. If that is a problem for Gradle, please do raise it with their team. They are a very responsive and friendly bunch.

Done.

src/main/scala/io/sdkman/changelogs/java/OpenJdkMigrations.scala
```diff
@@ -188,4 +188,31 @@ class OpenJdkMigrations {
       .foreach { version =>
         removeVersion("java", "16.ea.15-open", version.platform)
       }
+
+  @ChangeSet(
+    order = "077",
```
eddumelendez, line 193:
```suggestion
    order = "078",
```

src/main/scala/io/sdkman/changelogs/java/OpenJdkMigrations.scala
```diff
@@ -188,4 +188,31 @@ class OpenJdkMigrations {
...
+    id = "077-add_openjdk_java_16-loom-6",
```
eddumelendez, line 194:
```suggestion
    id = "078-add_openjdk_java_16-loom-6",
```

src/main/scala/io/sdkman/changelogs/java/OpenJdkMigrations.scala
```diff
@@ -188,4 +188,31 @@ class OpenJdkMigrations {
...
+    author = "soberich"
+  )
+  def migrate077(implicit db: MongoDatabase): Unit =
```
eddumelendez, line 197:
```suggestion
  def migrate078(implicit db: MongoDatabase): Unit =
```

src/main/scala/io/sdkman/changelogs/java/OpenJdkMigrations.scala
```diff
@@ -188,4 +188,31 @@ class OpenJdkMigrations {
...
+    Map(
+      Linux64    -> "openjdk-16-loom+6-105_linux-x64_bin.tar.gz",
+      MacOSX     -> "openjdk-16-loom+6-105_osx-x64_bin.tar.gz",
+      Windows    -> "openjdk-16-loom+6-105_windows-x64_bin.zip"
+    ).map {
+        case (platform, binary) =>
+          Version(
+            "java",
+            "16.0.6.lm-open", //has to be "1.2.3" e.i. numeric in order to Gradle auto detection work https://docs.gradle.org/nightly/userguide/toolchains.html#sec:auto_detection
```
eddumelendez, line 206:
```suggestion
            "16.0.6.lm-open",
```

src/main/scala/io/sdkman/changelogs/java/OpenJdkMigrations.scala
```diff
@@ -188,4 +188,31 @@ class OpenJdkMigrations {
...
+            s"https://download.java.net/java/early_access/loom/6/$binary",
+            platform,
+            Some(OpenJDK)
+          )
+      }
+      .toList
+      .validate()
+      .insert()
+      .foreach { version =>
+        removeVersion("java", "16.0.6.lm-open", version.platform)
```
eddumelendez, line 216: this is not needed

eddumelendez (changes requested):
@soberich thanks for your contribution. I have made some suggestion due to changelog `077` already exists

src/main/scala/io/sdkman/changelogs/java/OpenJdkMigrations.scala
marc0der: Mind if we change this to `16.ea.6.lm-open` as we previously discussed?

eddumelendez: suggesting that change too. goo catch @marc0der !

eddumelendez (approved):

eddumelendez:
@soberich thanks for your contribution! this is now in master

soberich:
@eddumelendez we've got a problem. `16.ea.6.lm-open` is already 15 symbols. when jdk build will have build number of two digits (which it already is), it is too long for sdkman .
`16.ea.14.lm-open` = 16 symbols

eddumelendez:
> 16.ea.6.lm-open is already 15 symbols

it is 17 characters now

PR #347, opened 2020-09-24
Add the SapMachine 15

poad:


src/main/scala/io/sdkman/changelogs/java/SapMachineMigrations.scala
```diff
@@ -83,4 +84,22 @@ class SapMachineMigrations {
...
+    author = "poad"
+  )
+  def migrate0012(implicit db: MongoDatabase) =
+    Map(
+      Linux64 -> "linux-x64_bin.tar.gz",
+      MacOSX  -> "osx-x64_bin.tar.gz",
+      Windows -> "windows-x64_bin.zip"
+    ).map {
+        case (platform, suffix) =>
+          toVersion(platform, suffix, "15")
```
eddumelendez, line 100:
```suggestion
          toVersion(platform, suffix, "15.0.0")
```

Also think we should not mix the identifier with the version used by the vendor

poad: Fixed it

eddumelendez (approved):

PR #349, opened 2020-09-24
Update SbtMigrations.scala

eed3si9n:


marc0der:
@eed3si9n thanks for the quick response! That's fixed now.

PR #367, opened 2020-10-12
Add Tomcat

larsmoil:


marc0der (changes requested):
For this to work, we first require the candidate to be created.

larsmoil:
I've added the candidate now. Is there any way to test things out locally before submitting a PR? It would be nice if I was able to use my local sdkman installation to verify my changes.

marc0der (approved):
Excellent job! Thank you!

PR #369, opened 2020-10-15
Update event type to pull_request_target for action.xml

eddumelendez:
Currently, when a PR is submitted by contributor from their own
fork, action is not running properly. In order to allow that
the action must use event `pull_request_target`.

### DIFF
--- a/.github/workflows/action.yml
+++ b/.github/workflows/action.yml
@@ -1,7 +1,7 @@
 ---
 # yamllint disable-line rule:truthy
 name: 'Auto Assign'
-on: pull_request
+on: pull_request_target
 
 jobs:
   add-reviews:


marc0der (approved):

.github/workflows/action.yml
```diff
@@ -1,7 +1,7 @@
 ---
 # yamllint disable-line rule:truthy
 name: 'Auto Assign'
-on: pull_request
```
marc0der, line 4: should it be for both `pull_request` and `pull_request_target`?

eddumelendez: according to the docs it will work for both purposes since it takes the base repository https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#pull_request_target

PR #372, opened 2020-10-18
upgrade url validator

marc0der:


.github/workflows/run.yml
```diff
@@ -0,0 +1,19 @@
+name: Run migration
+on: pull_request
```
eddumelendez, line 2:
```suggestion
on: pull_request_target
```

eddumelendez (approved):
run.yml already exists, isn't it? wondering why another is needed.

marc0der:
I think I must have added that in error @eddumelendez, I'll fix the PR.

marc0der:
That should be fixed now @eddumelendez

eddumelendez (approved):

PR #406, opened 2020-11-05
add adoptopenjdk-8u202

mattluce:


marc0der:
We'll start keeping historic versions very shortly.

PR #407, opened 2020-11-05
add adoptopenjdk 8 202

mattluce:


andrebrait:
What is the point, if `8.0.282.open-adpt` is already there?

mattluce:
Our team needs a very specific version of the jdk. However, there may be better ways than adding it to the global repository.

marc0der:
Apart from that, the URL is invalid and failing to build.

mattluce:
I think that failing URL is for a different library. It looks like other PRs may have had the same issue.

Anyway, I'll work on a different way of managing the specific version.

PR #412, opened 2020-11-09
Remove old OpenJDK 8 & 11 binaries

andrebrait:
Explaining the rationaly:

Given the new AdoptOpenJDK Upstream binaries are now the official binary builds of OpenJDK (with https://wiki.openjdk.java.net/display/jdk8u/Main even pointing at the GitHub-hosted binaries from AdoptOpenJDK's Upstream repositories), it's inconsistent to provide any releases under `*-open` at the same time we provide them under `*.open-adpt`.

Given we are providing OpenJDK 8 and OpenJDK 11 under `.open-adpt`, I want to remove the corresponding major releases from the `-open` options, as the ones from Java.net are outdated and superseded by those we now provide under AdoptOpenJDK's Upstream builds.

### DIFF
--- a/src/main/scala/io/sdkman/changelogs/java/OpenJdkMigrations.scala
+++ b/src/main/scala/io/sdkman/changelogs/java/OpenJdkMigrations.scala
@@ -208,4 +208,15 @@ class OpenJdkMigrations {
         removeVersion("java", "16.ea.22-open", version.platform)
       }
 
+  @ChangeSet(
+    order = "087",
+    id = "087-remove-java.net-openjdk-11-8-binaries",
+    author = "andrebrait"
+  )
+  def migrate087(implicit db: MongoDatabase): Unit =
+    List(Linux64, Windows)
+      .foreach { platform =>
+        removeVersion("java", "8.0.265-open", platform)
+        removeVersion("java", "11.0.2-open", platform)
+      }
 }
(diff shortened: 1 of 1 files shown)


andrebrait:
@eddumelendez I realize you had removed them and then re-added 11.0.2 under the Java.net section, but I think we can remove them given the now-official status of AdoptOpenJDK Upstream builds.

andrebrait:
CI failed, but not because of my changes. It's probably worth retrying?

eddumelendez:
@andrebrait thanks for sharing this with us. We are close to deliver a new feature in sdkman which will hide versions and just display the last ones. So, I think we can wait for it until then so people who relies on this can use it.

andrebrait:
@eddumelendez so both could exist in the DB and be accessible if needed but only the most recent would be shown by default?

Can that work between "variants"? Because right now one is under AdoptOpenJDK and the other is under Java.net.

eddumelendez:
> so both could exist in the DB and be accessible if needed but only the most recent would be shown by default?

Yes

> Can that work between "variants"? Because right now one is under AdoptOpenJDK and the other is under Java.net.

This is a special case. The reasoning behind this is that the upstream is delivered by AdoptOpenJDK instead of OpenJDK itself. So, we will be displaying the last one for each vendor. Well, this will be the behavior but can change.

@helpermethod @marc0der wdyt?

eddumelendez:
now that I have revisited the adoptopenjdk api and see openjdk as a vendor (dragonwell will be another vendor too) make sense to move the ones with `open-adpt` to OpenJDK and remove those old versions.

eddumelendez:
versions with `open-adpt` have been removed and nowadays we are supporting the hiding feature instead of removing. I will close the PR.

PR #432, opened 2020-11-19
Add Tomcat 8.5.60/9.0.40

eddumelendez:


marc0der:
I think we addressed these two versions in my PR. Can we close this @eddumelendez?

BTW, I noticed that we weren't using the `archives.apache.org` subdomain which has permalinks. The ones coming from the `downloads.apache.org` subdomain are deleted on every new version release, so we should consider standardizing on the others. WDYT?

PR #447, opened 2020-12-22
Add Scala 3.0.0-M2 and 3.0.0-M3

anaumidis:


marc0der:
Hi @anaumidis, thanks for contributing. dotty is distributed under its own candidate with a separate [migration file](https://github.com/sdkman/sdkman-db-migrations/blob/master/src/main/scala/io/sdkman/changelogs/DottyMigrations.scala).

anaumidis:
Hello @marc0der, yes, you are right.

But the fact that the new binaries are named `scala`, `scalac`, `scalad`, and the `3.0.0-M1` version is already available for this candidate made me believe that this is the right place for the migrations.

Of course, if that's not the case I'll happily move the migrations to the Dotty candidate. :slightly_smiling_face:

marc0der:
Ha! That makes perfect sense then. Thanks for contributing!

marc0der (approved):

PR #450, opened 2020-12-28
Fix WINDOWS_X64 and add MAC_ARM64

eddumelendez:


marc0der:
This should be as it was. I've fixed it in the broker service.

marc0der:
Closing this as they issue has been resolved.

PR #454, opened 2021-01-16
Hide 16.ea.9.lm-open

eddumelendez:


marc0der (approved):

soberich:
there is no description.
Why is it hidden?

marc0der:
@soberich versions are usually hidden when superseded. they are still available, just no longer listed.

@eddumelendez I see we no longer list a `16.ea` `lm-open` release. Is this intentional?

eddumelendez:
latest versions are 17-ea for loom and panama

soberich:
@marc0der @eddumelendez 
Since when `superseded` implies major version bump? Especially for JDK..
I don't know what was the motivation to remove/hide. If the idea was that "since it is a (sub)project in Openjdk we can bump major version (sub)project's build should be smart enough handle it automagicaly" - then well, this is not true and the assumption itself I understand sounds a bit non-realistic. It's the same as for example, remove java 8 binaries and say everyone please use java 9.
Incompatible changes hit you there.
Let's see it as a mistake. Could you please put it back

soberich:
@marc0der @eddumelendez ping

eddumelendez:
@soberich as you may see the list of java candidates in general is getting long. There was an approach at the beginning which was removing the version and keep the latest. After that, we decide to keep the versions and instead just hiding it? why, simple reason the list is growing. There is another reason about this which is that we will be displaying just latest versions of LTS such as java 8 and 11 for each vendor and the lastest version release which as of today it is 16. Why hiding 15? because there will not be more binaries for that version. Just to be clear, we are not promoting to use latest versions. So, if you are have scripts using 15.0.2-open for example, this should be still working.

PR #458, opened 2021-01-20
Update Bellsoft Liberica JDK version to 8u282, 11.0.10 and 15.0.2

poad:
## Added
- 8u282
- 11.0.10
- 15.0.2
- 8u282 with FX
- 11.0.10 with FX
- 15.0.2 with FX

## Removed
 - 8u275b1
 - 8u275b1 with FX
 - 11.0.9+1
 - 11.0.9+1 with FX
 - 15.0.1
 - 15.0.1 with FX


eddumelendez:
@poad thanks for the PR but this is not needed. BellSoft Liberica versions are already available via sdkman.

eddumelendez:
FYI: Manual migrations are not needed anymore and nowadays we are not removing versions instead we are hiding them. see https://github.com/sdkman/sdkman-db-migrations/pull/456

PR #461, opened 2021-01-23
Hide java versions

eddumelendez:
Hide old versions and no lts versions for the different vendors. Just keep last no lts version 15.


marc0der:
Wondering if can't automate this somehow. I was wondering if we should introduce an API endpoint for the purpose of hiding. wdyt, less work for you right?

marc0der (approved):

PR #475, opened 2021-02-28
webtau initial registration

MykolaGolubyev:


src/main/scala/io/sdkman/changelogs/WebtauMigrations.scala
```diff
@@ -0,0 +1,35 @@
...
+    order = "001",
+    id = "001_add_webtau_candidate",
+    author = "MykolaGolubyev"
+  )
+  def migration001(implicit db: MongoDatabase) =
+    Candidate(
+      candidate = "webtau",
+      name = "webtau",
+      description =
+        "webtau - web test automation. test across REST-API/Graph QL/Browser/Database/CLI/Business-Logic",
```
marc0der, line 18: This is the description that appears on our website, and also when viewing the `sdk list` command. Can you please make this a little more descriptive and let it be a real sentence/paragraphl?

src/main/scala/io/sdkman/changelogs/WebtauMigrations.scala
```diff
@@ -0,0 +1,35 @@
...
+      websiteUrl = "https://github.com/testingisdocumenting/webtau"
+    ).insert()
+
+  @ChangeSet(
```
marc0der, line 22: We don't need this migration as you will be doing your first version release with the API.

marc0der (changes requested):

src/main/scala/io/sdkman/changelogs/WebtauMigrations.scala
MykolaGolubyev: I copied it as is from the webtau repo. Do you have any suggestions how can I improve it? 
Tool/API allows to write tests that touch one or multiple layers within one test or suite; generate rich report and other goodies. 
What is the size limit?

marc0der: There is no real size limit, but it is probably best to look at what other vendors are doing for general length and tone.

MykolaGolubyev: updated description

marc0der (approved):

marc0der:
Thanks for contributing!

PR #481, opened 2021-03-08
Mandrel: Remove old versions and replace 21.0 and 20.3 with respins

zakkak:
Removes obsolete 20.1.0.3 and 20.2.0.0.
Replaces 20.3.1.2 and 21.0.0.0 with respins (same code but with different versioning scheme) due to https://github.com/graalvm/mandrel/issues/218

### DIFF
--- a/src/main/scala/io/sdkman/changelogs/java/MandrelMigrations.scala
+++ b/src/main/scala/io/sdkman/changelogs/java/MandrelMigrations.scala
@@ -2,7 +2,14 @@ package io.sdkman.changelogs.java
 
 import com.github.mongobee.changeset.{ChangeLog, ChangeSet}
 import com.mongodb.client.MongoDatabase
-import io.sdkman.changelogs.{Linux64, Windows, Mandrel, Version}
+import io.sdkman.changelogs.{
+  Linux64,
+  Windows,
+  Mandrel,
+  Version,
+  removeVersion,
+  hideVersion
+}
 
 @ChangeLog(order = "035")
 class MandrelMigrations {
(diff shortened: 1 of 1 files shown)


zakkak:
Note: I was not able to test locally due to "Bandwidth limit exceeded" error on archive.apache.org

eddumelendez:
Hi @zakkak, thank you so much for the PR! We currently are not removing versions. Instead, we are hiding them. See https://github.com/sdkman/sdkman-db-migrations/commit/f28b2ca29953214aba9c78bf39c4e699cbc734e6

zakkak:
OK, so I guess I should hide 20.1.0.3 and 20.2.0.0, and remove and add again 21.0.0.0 and 20.3.1.2 (to force the update). Correct?

However, now that I think about it if I remove and re-add a version users with that version already installed would also have to remove the corresponding version and re-install it. The problem here is that the only difference between the new and the old binaries is that in the former the version appears as `20.3.1.2-Final` while in the latter it appears as `20.3.1.2.Final`. I guess it should be OK to just wait for the next minor release (coming in April).

eddumelendez:
if nothing have change and it's a filename issue we can skip it. otherwise, it can be removed.

zakkak:
@eddumelendez can you please review this?
It will help close https://github.com/quarkusio/quarkus/issues/15646

eddumelendez (approved):

eddumelendez:
@zakkak sorry for the delay. this is already merged.

zakkak:
> @zakkak sorry for the delay. this is already merged.

No worries :)

Thank you Eddu!

PR #483, opened 2021-03-15
Fix urls for AdoptOpenJDK 8.0.272

helpermethod:
Except for the Linux 64 Arm platform, all other platforms point to the 8.0.275 release.

This PR fixes the broken URLs.


marc0der (approved):
:rocket:

marc0der:
@helpermethod looks like this caused the build to fail. https://travis-ci.org/github/sdkman/sdkman-db-migrations

PR #485, opened 2021-03-20
Hide Java 15 and 16 EA

eddumelendez:


src/main/scala/io/sdkman/changelogs/java/AdoptOpenJdkMigrations.scala
```diff
@@ -1182,4 +1182,16 @@ class AdoptOpenJdkMigrations {
...
+  @ChangeSet(
+    order = "057",
+    id = "057-hide-java-versions",
+    author = "eddumelendez"
+  )
+  def migrate057(implicit db: MongoDatabase): Unit =
+    Seq(
+      "15.0.2.j9-adpt",
+      "15.0.2.hs-adpt",
+      "8.0.272.hs-adpt"
```
helpermethod, line 1195: Thanks, forgot about hiding the version after fixing the URLs :).

helpermethod (approved):
LGTM!

renannprado:
@helpermethod @eddumelendez what was the reason to hide Java 15? Consider that people develop software that run on specific versions of Java therefore I would expect *at least* the latest release of any java version to be available through sdkman.

eddumelendez:
because the next non LTS version become available, in this case java 16. sdkman will display LTS versions and the last non LTS version. But, you still can use 15.0.2-open, sdkman still provide those binaries

renannprado:
@eddumelendez how could I get to know what to type in order to install the version if it's hidden? is there a flag to unhide the versions?

marc0der:
Hi @renannprado, we simply don't have the space to display all historic versions, so we hide non LTS (EOL) releases.

There is also a security implication by staying on EOL versions. If you are still on such a version, I suggest you move onto a newer major version asap. These non LTS versions reach end of life the moment a subsequent version is released, so they are effectively abandoned by the community.

renannprado:
> Hi @renannprado, we simply don't have the space to display all historic versions, so we hide non LTS (EOL) releases.
> 
> There is also a security implication by staying on EOL versions. If you are still on such a version, I suggest you move onto a newer major version asap. These non LTS versions reach end of life the moment a subsequent version is released, so they are effectively abandoned by the community.

@marc0der I understand what you're saying, however IMHO upgrading the version of java is a choice the users of sdkman should make and not the owners of the tool should push for. As you know, for example, upgrading java is not something you can do always easily like that if the framework depends or library you use depends on a specific version even though nowadays things are smoother in the new java versions.

Issue #443, opened 2020-12-17
Add support for older Scala versions


Eugenio-Liso:
Hi all!

Right now, SDKMAN! supports Scala SDK versions starting from 2.11.0. Since there are a lot of projects out there that are still using older versions, is that possible to add support for 2.10.X?  

I think that should be enough, since the latest 2.9.X version (2.9.3) has been tagged in 2013, so it's pretty old.

marc0der:
Hi @Eugenio-Liso, the more versions the merrier. Please feel free to add them by raising a PR on this project.

PR #487, opened 2021-03-23
Add support for older Scala versions

Eugenio-Liso:
Pull Request opened for issue: https://github.com/sdkman/sdkman-db-migrations/issues/443

This PR adds all versions of Scala 2.10.x in the `ScalaMigrations` class.

### DIFF
--- a/src/main/scala/io/sdkman/changelogs/ScalaMigrations.scala
+++ b/src/main/scala/io/sdkman/changelogs/ScalaMigrations.scala
@@ -263,98 +263,31 @@ class ScalaMigrations {
 
   @ChangeSet(
     order = "020",
-    id = "020_add_scala_2.10.1",
+    id = "020_add_scala_2.10.x",
     author = "eliso"
   )
-  def migration020(implicit db: MongoDatabase): Unit =
-    Version(
-      candidate = "scala",
-      version = "2.10.1",
-      url = "https://downloads.lightbend.com/scala/2.10.1/scala-2.10.1.zip"
-    ).validate()
-List("2.10.1", "2.10.2", "2.10.3", .....).map { version =>
-  Version(
-      candidate = "scala",
-      version = version,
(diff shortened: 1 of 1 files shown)


src/main/scala/io/sdkman/changelogs/ScalaMigrations.scala
```diff
@@ -260,4 +260,95 @@ class ScalaMigrations {
...
+    id = "020_add_scala_2.10.1",
+    author = "eliso"
+  )
+  def migration020(implicit db: MongoDatabase): Unit =
+    Version(
+      candidate = "scala",
+      version = "2.10.1",
+      url = "https://downloads.lightbend.com/scala/2.10.1/scala-2.10.1.zip"
+    ).validate()
+      .insert()
```
marc0der, line 275: How about trying some like this instead?

```suggestion
List("2.10.1", "2.10.2", "2.10.3", .....).map { version =>
  Version(
      candidate = "scala",
      version = version,
      url = "https://downloads.lightbend.com/scala/${version}/scala-${version}.zip"
  )
}.validate().insert()
```

marc0der:
Hi @Eugenio-Liso,

Thanks for your contribution!

That will work, but you can probably simplify it a lot with what I've suggested.

marc0der:
Don't know what happened, but it seems like it committed my suggestion! Sorry!

src/main/scala/io/sdkman/changelogs/ScalaMigrations.scala
Eugenio-Liso: Thanks for you comment!

I've actually generated those methods by iterating on the versions - as you pointed out - and then pasted them in the class.
I totally agree with this. I will change that.

Eugenio-Liso: Changes pushed. Let me know!

marc0der (approved):

marc0der:
Thanks for contributing!

PR #490, opened 2021-04-03
Add AdoptOpenJDK Hotspot 222

henri-tremblay:


marc0der (changes requested):
Hi @henri-tremblay, 

Thanks for this PR!

It seems like the build is broken due to an orphaned binary. I'll be happy to merge if the build is green.

Also, we need to set this version as explicitly hidden.

henri-tremblay:
The build is indeed broken. But that doesn't seem caused by me. It's about a missing Loom binary.

henri-tremblay:
I don't know how you fix that normally. There are 3 early access versions that are not available anymore, I tried to remove them in a new change set but it doesn't work because it fails on the old charsets. Shall I delete the 3 problematic change sets? Build passes if I do so.

henri-tremblay:
I look at other PRs and it seems to be the way to go. So I've deleted the 3 obsolete change sets

henri-tremblay:
Weird. The failure is now a false negative. I don't see how I could re-run the checks.

marc0der:
I usually run the migration locally as suggested in the README and remove any defunct migrations in a second commit.

marc0der:
Never mind, I see that is exactly what you did 😀

marc0der (approved):
LGTM.

PR #513, opened 2021-05-23
Update default Java candidate

nickebbitt:
Sets 11.0.11.hs-adpt as the default Java candidate. 11.0.11 is the most recent patch release for the current LTS.


helpermethod (approved):
LGTM!

PR #525, opened 2021-07-15
Adding new candidate ActiveMQ (classic)

xshyamx:
Adding new candidate for Apache ActiveMQ (classic)


marc0der (changes requested):
Thanks for the PR!

One thing, I'd prefer the absolute archive URLs for the artifacts. These dynamic Apache links have caused us many headaches in the past. If you could change this I'd be happy to merge.

xshyamx:
The archive urls were pretty slow when I was testing so, used the dynamic mirror url. Have reverted to use the apache archive urls

marc0der (approved):

marc0der:
Thanks for contributing!

PR #536, opened 2021-08-15
Adding Spark and Flink minor versions

ChethanUK:


marc0der (commented):
Unfortunately, db migrations are always immutable. If you want to add new versions, you need to add a new changeset.

Another thing about our DSL is that `Version` has a method to set it as the default candidate version.

Issue #540, opened 2021-08-23
Add checksums to versions


hgeraldino:
Adding this issue to track the work required on the db-migrations module to support https://github.com/sdkman/sdkman-cli/issues/945

marc0der:
Happy that this is happening!

I'd suggest we chat on Slack prior to kicking of individual pieces of work. That way I can guide you in the optimal way of doing things in the backend, which should hopefully save you some time in implementing this. You've also probably noticed that we're big on testing, so we expect most pieces of functionality to be covered somehow.

PR #541, opened 2021-08-23
Add checksums

hgeraldino:
Store checksum info for downloadable artifacts (see #540)

This PR adds 2 new fields to the `versions` document to store the SHA algorithm and hash for each downloadable artifact.  It also includes a migration step to update the checksums for Maven artifacts. If you guys are ok with this general direction I'll add checksums for the rest of the candidates and update the README.md

This change is needed before submitting PRs to update the broker, cli and vendor-release repositories.

### DIFF
--- a/src/main/scala/io/sdkman/changelogs/package.scala
+++ b/src/main/scala/io/sdkman/changelogs/package.scala
@@ -28,9 +28,9 @@ package object changelogs {
     new Validator[Version] with UrlValidation with LazyLogging {
 
       override def validVersion(v: Version): Unit = v match {
-        case Version("java", version, _, _, _) =>
+        case Version("java", version, _, _, _, _, _) =>
           if (version.length > 17) throw exception(version)
-        case Version(_, version, _, _, _) =>
+        case Version(_, version, _, _, _, _, _) =>
           if (version.length > 15) throw exception(version)
       }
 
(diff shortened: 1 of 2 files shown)


src/main/scala/io/sdkman/changelogs/package.scala
```diff
@@ -200,7 +224,9 @@ package object changelogs {
       version: String,
       url: String,
       platform: Platform = Universal,
-      vendor: Option[Vendor] = None
+      vendor: Option[Vendor] = None,
+      hashAlgorithm: Option[HashAlgorithm] = None,
+      checksum: Option[String] = None
```
marc0der, line 229: I'm not sure if that entirely makes sense. Can a checksum exist without an algorithm? These two together should probably have a case class of their own, and that should be an option.
Something like:
```
case class Checksum(algorithm: HashAlgorithm, value: String)
...
checksum: Option[Checksum]
```

marc0der (commented):
Looking good! I suggest some changes to the modelling (which will affect how this is persisted too). I also thought about if we wanted to store multiple checksums for a version, say an md5 and sha256. Would a list of `Checksum`s be better than the option?

One last thing, can you please try to rebase your feature branch rather than merge from master? As it stands now, it will cause multiple merge commits in our timeline on master.

src/main/scala/io/sdkman/changelogs/package.scala
stewSquared: Agreed. But if this is slowing down the pull request, I'd be to contribute an edit.

marc0der: @stewSquared perhaps you aren't getting the context here, but you are looking at the database migration project. This will be the last thing to be implemented. It is also the least important of the lot.

We are driving this feature from the client side, and are currently focusing on the CLI. After that we will do the crucial work in the broker, followed by all the other auxillary services.

At the moment, everything works on the client, with the exception of one small bug that we've encountered with cached archives. The fix will be released soon. Checksums will be released into the wild when we think it's ready, not a moment sooner.

PR #557, opened 2021-11-14
Pierrot candidate and initial version

musketyr:
This PR adds a candidate and the latest version of Pierrot CLI to SDKMAN.

### DIFF
--- a/src/main/scala/io/sdkman/changelogs/PierrotMigrations.scala
+++ b/src/main/scala/io/sdkman/changelogs/PierrotMigrations.scala
@@ -1,0 +1,51 @@
+package io.sdkman.changelogs
+
+import com.github.mongobee.changeset.{ChangeLog, ChangeSet}
+import com.mongodb.client.MongoDatabase
+
+@ChangeLog(order = "063")
+class PierrotMigrations {
+  @ChangeSet(
+    order = "001",
+    id = "001_add_pierrot_candidate",
+    author = "musketyr"
+  )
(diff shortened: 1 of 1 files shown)


src/main/scala/io/sdkman/changelogs/PierrotMigrations.scala
```diff
@@ -0,0 +1,51 @@
...
+      websiteUrl = "https://agorapulse.github.io/pierrot/",
+      distribution = "PLATFORM_SPECIFIC"
+    ).insert()
+
+  @ChangeSet(
+    order = "002",
+    id = "002-add_pierrot_0_9_8",
+    author = "musketyr"
+  )
+  def migration002(implicit db: MongoDatabase) = {
```
marc0der, line 28: We probably don't want versions here, because you will be doing this with the API.

musketyr: can we do just the initial here to have something to start with before I setup the integration?

marc0der: SDKMAN will report it on the list command and on the website as "coming soon", but I'm happy to go ahead with these versions if you want.

musketyr: yes, please :-) it might take some time to implement the integration so I would like to have some starter version

PR #560, opened 2021-11-25
Update Tomcat versions to the latest

chubbard:
Updated the versions of Apache Tomcat to the latest versions for series 7 -> 7.0.109, 8.5 -> 8.5.73, 9 -> 9.0.55, and 10 -> 10.0.13.  Also updated the default candidate from 9 to 10 since 10 has been officially released now.

### DIFF
--- a/src/main/scala/io/sdkman/changelogs/TomcatMigration.scala
+++ b/src/main/scala/io/sdkman/changelogs/TomcatMigration.scala
@@ -29,10 +29,10 @@ class TomcatMigration {
   def migration007(implicit db: MongoDatabase) = {
     removeAllVersions("tomcat")
     List(
-      "7"  -> "7.0.106",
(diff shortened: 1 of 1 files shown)


eddumelendez (changes requested):
new migration should be added. updating existing one will not apply the change.

ahmadshalabi:
> new migration should be added. updating existing one will not apply the change.

@eddumelendez, I did the necessary change in this [pull request](https://github.com/sdkman/sdkman-db-migrations/pull/573)

marc0der:
@ahmadshalabi database migrations are always immutable. you need to add a new migration below the existing one (with a new changelog number) for it to take effect.

eddumelendez:
closing because those and even greater versions are already available via sdkman

PR #569, opened 2022-01-03
Add Maven 4 Alpha 4 snapshot version

helpermethod:
It's a great way to let people test out a *very* early version of Maven 4 and the brand new Maven Wrapper, which requires Maven 4.


src/main/scala/io/sdkman/changelogs/MavenMigration.scala
```diff
@@ -105,4 +105,14 @@ class MavenMigration {
       .asCandidateDefault()
   }
 
+  @ChangeSet(order = "009", id = "009-add_maven_4.0.0-alpha-1-SNAPSHOT", author = "helpermethod")
+  def migration009(implicit db: MongoDatabase) = {
+    Version(
+      "maven",
+      "4.0.0-alpha-1-SNAPSHOT",
```
marc0der, line 112: Mind if we drop `-SNAPSHOT` from the version? I think it might exceed column bounds and it doesn't really add anything to the uniqueness of the version.

helpermethod: Good point!

marc0der (approved):

PR #573, opened 2022-01-04
Update Tomcat versions to the latest #560

ahmadshalabi:
I update tomcat versions to the latest. Also, I set the default to 10.0.14.

### DIFF
--- a/src/main/scala/io/sdkman/changelogs/TomcatMigration.scala
+++ b/src/main/scala/io/sdkman/changelogs/TomcatMigration.scala
@@ -31,4 +31,31 @@ class TomcatMigration {
       .insert()
     setCandidateDefault("tomcat", "9.0.40")
   }
+
+  @ChangeSet(
+    order = "008",
+    id = "008-update_tomcat_versions",
+    author = "ahmadshalabi"
+  )
+  def migration008(implicit db: MongoDatabase): Document = {
+    removeAllVersions("tomcat")
+    List(
+      "7"  -> "7.0.109",
+      "8"  -> "8.5.73",
+      "9"  -> "9.0.56",
+      "10" -> "10.0.14"
+    ).map {
+        case (series: String, version: String) =>
+          Version(
+            candidate = "tomcat",
+            version = version,
+            url =
+              s"https://archive.apache.org/dist/tomcat/tomcat-$series/v$version/bin/apache-tomcat-$version.zip"
+          )
+      }
+      .validate()
+      .insert()
+    setCandidateDefault("tomcat", "10.0.14")
+  }
+
 }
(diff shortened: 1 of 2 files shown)


ahmadshalabi:
`Error: Exception in thread "main" com.github.mongobee.exception.MongobeeException: Invalid url: https://repository.apache.org/content/repositories/snapshots/org/apache/maven/apache-maven/4.0.0-alpha-1-SNAPSHOT/apache-maven-4.0.0-alpha-1-20211230.155705-237-bin.zip`

The issue is the current maven-4.0.0-alpha-1 no longer exists in the apache repository and this will cause a failure when someone wants to try the alpha version.


CC: @eddumelendez @marc0der

src/main/scala/io/sdkman/changelogs/TomcatMigration.scala
```diff
@@ -31,4 +31,31 @@ class TomcatMigration {
...
     setCandidateDefault("tomcat", "9.0.40")
   }
+
+  @ChangeSet(
+    order = "008",
+    id = "008-update_tomcat_versions",
+    author = "ahmadshalabi"
+  )
+  def migration008(implicit db: MongoDatabase): Document = {
+    removeAllVersions("tomcat")
```
marc0der, line 41: Why not keep the legacy versions? SDKMAN is specifically for allowing parallel version of any SDK.

ahmadshalabi: Thanks for reminding me to keep the legacy versions.

marc0der:
If the maven team don't understand that resources on the web should be immutable, I don't see the point of hosting these 4 alpha binaries at all. We'll continually be playing a wack-a-mole game trying to keep them up to date. wdyt @helpermethod?

helpermethod:
@marc0der It was probably a bad idea from my side to add them. I guess we should just remove them and wait for a stable release.

marc0der:
@helpermethod I thought your idea was great, I just wasn't expecting how they would tamper with web resources. Especially since they are maven 😄

marc0der:
@helpermethod I thought your idea was great, I just wasn't expecting how they would tamper with web resources. Especially since they are maven 😄

src/main/scala/io/sdkman/changelogs/TomcatMigration.scala
ahmadshalabi: I did the change and added the last 10.1.0 version (10.1.0-M8).

marc0der (approved):

marc0der:
We might need to remove the maven 4 migration to get the build green before merging.

ahmadshalabi:
@marc0der, I opened #1033 to track the deletion of the maven 4 version.

marc0der:
Thanks for your contribution @ahmadshalabi

ahmadshalabi:
> Thanks for your contribution @ahmadshalabi

Most welcome @marc0der.

PR #578, opened 2022-01-18
Add VisualVM 2.1.2

eddumelendez:


src/main/scala/io/sdkman/changelogs/VisualVmMigrations.scala
```diff
@@ -6,15 +6,15 @@ import com.mongodb.client.MongoDatabase
...
 class VisualVmMigrations {
   @ChangeSet(
-    order = "012",
-    id = "012-add_visualvm_2_1_1",
+    order = "013",
+    id = "013-add_visualvm_2_1_2",
     author = "eddumelendez"
   )
-  def migration012(implicit db: MongoDatabase): Unit = {
+  def migration(implicit db: MongoDatabase): Unit = {
```
helpermethod, line 13: should probably be `def migration013`, no? In case there will be more migrations in the future.

helpermethod (dismissed):

src/main/scala/io/sdkman/changelogs/VisualVmMigrations.scala
eddumelendez: I think the current approach is to keep just the latest one. last change was removing oldest ones by @marc0der

helpermethod: Hi Eddu,

If that's the case the MR is fine. Sorry :).

helpermethod (approved):

src/main/scala/io/sdkman/changelogs/VisualVmMigrations.scala
eddumelendez: no worries, thanks!

PR #582, opened 2022-01-28
updating new jmeter versions and setting new default

xshyamx:
Update with new JMeter versions and set the default to the latest version

### DIFF
--- a/src/main/scala/io/sdkman/changelogs/JMeterMigration.scala
+++ b/src/main/scala/io/sdkman/changelogs/JMeterMigration.scala
@@ -8,6 +8,21 @@ import org.bson.Document
 class JMeterMigration {
 
   @ChangeSet(
+    order = "001",
+    id = "001-add_jmeter_candidate",
+    author = "xshyamx"
+  )
+  def migration001(implicit db: MongoDatabase): Candidate = {
+    Candidate(
+      candidate = "jmeter",
+      name = "Apache JMeter",
+      description =
+        "The Apache JMeter™ application is open source software, a 100% pure Java application designed to load test functional behavior and measure performance. It was originally designed for testing Web Applications but has since expanded to other test functions.",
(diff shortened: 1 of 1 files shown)


src/main/scala/io/sdkman/changelogs/JMeterMigration.scala
```diff
@@ -2,21 +2,41 @@ package io.sdkman.changelogs
 
 import com.github.mongobee.changeset.{ChangeLog, ChangeSet}
 import com.mongodb.client.MongoDatabase
+import org.bson.Document
 
 @ChangeLog(order = "051")
 class JMeterMigration {
+  val CandidateName  = "jmeter"
+  val DefaultVersion = "5.4.3"
```
marc0der, line 10: You can probably ditch these constants, as they will only ever be used in one place and make things more confusing to grok.

xshyamx: Removed

src/main/scala/io/sdkman/changelogs/JMeterMigration.scala
```diff
@@ -2,21 +2,39 @@ package io.sdkman.changelogs
...
+            version = version,
+            url =
+              s"https://archive.apache.org/dist/jmeter/binaries/apache-jmeter-$version.zip"
+          )
+      )
+      .validate()
+      .insert()
+  }
   @ChangeSet(
-    order = "001",
```
marc0der, line 9: Sorry to be a pain, but let's keep the main candidate migration here for reference like we've done everywhere else. We're only removing version migrations for brevity.

xshyamx: No problem. I understand the reasoning behind :). Updated the PR

marc0der (approved):
Thanks for contributing!

PR #594, opened 2022-04-14
Add Maven versions 3.3.3 to 3.0.4

JohnLBergqvist:
These historical Maven 3.x versions are also available on https://archive.apache.org/dist/maven/maven-3, and I feel it would be useful to also have these available to pick from, alongside the prior versions that are already there.


marc0der (approved):
LGTM! Thanks for contributing!

PR #598, opened 2022-06-11
Add Maven 3.8.6

aalmiray:


src/main/scala/io/sdkman/changelogs/MavenMigration.scala
```diff
@@ -5,15 +5,25 @@ import com.mongodb.client.MongoDatabase
...
+      "3.8.6",
+      "https://archive.apache.org/dist/maven/maven-3/3.8.6/binaries/apache-maven-3.8.6-bin.zip"
+    ).validate()
+      .insert()
+      .asCandidateDefault()
+  }
+  
   @ChangeSet(order = "009", id = "009-add_maven_3.8.5", author = "aalmiray")
-  def migration008(implicit db: MongoDatabase): Unit = {
+  def migration009(implicit db: MongoDatabase): Unit = {
```
marc0der, line 20: We can probably remove all the old unused migrations.

aalmiray: Leaving only the last one for `3.8.6`? I'll do it in a jiffy.

eddumelendez (approved):

PR #603, opened 2022-07-16
updating latest mybatis version 3.3.11

xshyamx:
Updating latest mybatis version 3.3.11 and setting it as the default


eddumelendez (approved):

eddumelendez:
thanks @xshyamx! it should be available in a couple of minutes

PR #608, opened 2022-08-17
update toolkit default version and deal with error

sanshengshui:


sanshengshui:
There is an error in the master code, the new PR solves this problem and modifies the version

marc0der (changes requested):
Database migrations are immutable, so you need to raise a PR with a new changelog for this to take effect. Also, the build seems broken.

sanshengshui:
got it!

PR #611, opened 2022-09-13
Add protoc (protobuf compiler) candidate

manojshr:


marc0der (commented):
Hi @manojshr, I don't really feel that protoc is well suited for SDKMAN. We focus on JVM related SDKs, and I don't think protoc falls into that category. It feels like more of a general purpose tool that can be installed by your package manager.

manojshr:
> Hi @manojshr, I don't really feel that protoc is well suited for SDKMAN. We focus on JVM related SDKs, and I don't think protoc falls into that category. It feels like more of a general purpose tool that can be installed by your package manager.

@marc0der I agree with you, but my intention behind adding this tool is to make sdkman a rich CLI manager which can be a suitable tool to manage all the widely used command line tools. Is there any plan to move sdkman in this direction? I will be happy to contribute.

marc0der:
Hey @manojshr, we have no plans to go in this direction right now. It is also important to note that SDKMAN is not a package, but an SDK manager. It is also not for general purpose SDK, but specifically for the JVM. We certainly don't want to compete with all the great package managers already out there. They are great, but we fulfil a different and niche group.

manojshr:
Thanks @marc0der for clarification, Closing this PR.

PR #605, opened 2022-07-20
Adding Spark 3.2 and Flink latest version

chethanuk-plutoflume:

PR #614, opened 2022-11-16
Add Spark 3.2.2, 3.3.0, and 3.3.1

cphbrt:
This is my first contribution to this repository, so please advise on any style or process mistakes.

An existing PR is open for 3.3.0 but is stalled: https://github.com/sdkman/sdkman-db-migrations/pull/605

The `./gradlew clean run` completes successfully on my machine with this PR's changes, but I had to use the command `docker run -d -p 27017:27017 --name=mongo arm64v8/mongo:4.0.28` to setup Mongo on my machine, which differs from the example on the README because I am on Apple silicon and Docker Hub doesn't have a mongo `3.2` for `arm64`.

I did _not_ test if the `sdk install spark 3.3.1` will succeed with this PR's changes, because I am not sure how to hook up my local SDKMAN with the mongoDB container...

So, tl;dr, I followed the patterns I see in Git closely. Happy to change, test, or add anything you advise.

Thanks!

PS: I got the new Spark version URLs from here: https://archive.apache.org/dist/spark/

You may notice that the end of the URL switched from `3.2.tgz` to `3.tgz` at Spark `3.3.X`. It looks comparably harmless to the change from `2.7.tgz` to `3.2.tgz` that occurred several commits ago in this repo.

### DIFF
--- a/src/main/scala/io/sdkman/changelogs/SparkMigrations.scala
+++ b/src/main/scala/io/sdkman/changelogs/SparkMigrations.scala
@@ -6,31 +6,45 @@ import com.mongodb.client.MongoDatabase
 @ChangeLog(order = "007")
 class SparkMigrations {
   @ChangeSet(
-    order = "017",
-    id = "016-add_spark_3.2.0",
-    author = "ChethanUK"
+    order = "020",
+    id = "020-add_spark_3.2.2",
+    author = "cphbrt"
   )
-  def migration017(implicit db: MongoDatabase): Unit = {
+  def migration020(implicit db: MongoDatabase) = {
     Version(
       "spark",
-      "3.2.0",
-      "https://archive.apache.org/dist/spark/spark-3.2.0/spark-3.2.0-bin-hadoop3.2.tgz"
+      "3.2.2",
+      "https://archive.apache.org/dist/spark/spark-3.2.2/spark-3.2.2-bin-hadoop3.2.tgz"
     ).validate()
       .insert()
-      .asCandidateDefault()
   }
 
   @ChangeSet(
-    order = "019",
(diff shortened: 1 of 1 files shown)


GurRonenExplorium:
Would love this as well! 👍

hgeraldino (changes requested):
Thanks for your contribution @cphbrt 

The methods on these migration classes should be treated as immutable; that is, instead of editing the existing migrations (order 17, 19 and so on) you should instead add new ones, following the same sequences.

eddumelendez (approved):
LGTM

cphbrt:
Looking specifically at `git log src/main/scala/io/sdkman/changelogs/SparkMigrations.scala`, it appears that commit 801f6d4ae5ed9a5fd2de4713cc19a76c5b432f06 by @marc0der was the biggest step in removing old migrations. That pattern was continued in 4d293e058c11e0e2b1f1ad4db3ad69fea74bd3ec by @chethanuk-plutoflume. However, in the currently open PR from @chethanuk-plutoflume (https://github.com/sdkman/sdkman-db-migrations/pull/605), @chethanuk-plutoflume left the previous migration present and [suggested](https://github.com/sdkman/sdkman-db-migrations/pull/605#issuecomment-1216581515) we remove it in a subsequent PR.

That appears to be the history. I don't know enough about the setup to determine if it was a _good_ idea to remove old migrations.

I'll wait to make changes until y'all talk through a plan. I could restore all the migrations removed from `src/main/scala/io/sdkman/changelogs/SparkMigrations.scala` in 801f6d4ae5ed9a5fd2de4713cc19a76c5b432f06 and 4d293e058c11e0e2b1f1ad4db3ad69fea74bd3ec if desired.

PR #646, opened 2023-05-15
Adds Detekt as a new candidate

helpermethod:
See https://github.com/detekt/detekt/issues/6086

### DIFF
--- a/src/main/scala/io/sdkman/changelogs/DetektMigrations.scala
+++ b/src/main/scala/io/sdkman/changelogs/DetektMigrations.scala
@@ -1,0 +1,26 @@
+package io.sdkman.changelogs
+
+import com.github.mongobee.changeset.{ChangeLog, ChangeSet}
+import com.mongodb.client.MongoDatabase
+
+@ChangeLog(order = "079")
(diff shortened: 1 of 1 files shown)


marc0der (approved):

src/main/scala/io/sdkman/changelogs/DetektMigrations.scala
```diff
@@ -0,0 +1,26 @@
+package io.sdkman.changelogs
+
+import com.github.mongobee.changeset.{ChangeLog, ChangeSet}
+import com.mongodb.client.MongoDatabase
+
+@ChangeLog(order = "079")
```
eddumelendez, line 6: I think it should be 080. I just noticed it

helpermethod: That's probably bc the PR was open for too long. Can we still change it?

marc0der: I wouldn't bother too much because the migration succeeded.

PR #647, opened 2023-05-15
Add new karaf versions and update default

FColinet:


FColinet:
Why ?
 INFO com.github.mongobee.Mongobee - Mongobee is releasing process lock.

marc0der (approved):

PR #651, opened 2023-06-09
GCN candidate

ezzarghili:
Graal Cloud Native GCN is an open source project (https://github.com/oracle/gcn) for cloud-native microservices.
First version we are submitting as PR, we will follow the vendor API process for future releases.

### DIFF
--- a/src/main/scala/io/sdkman/changelogs/GcnMigrations.scala
+++ b/src/main/scala/io/sdkman/changelogs/GcnMigrations.scala
@@ -22,7 +22,7 @@ class MicronautMigrations {
     ).insert()
   }
 
-  def migrate007(implicit db: MongoDatabase): Unit = {
+  def migrate002(implicit db: MongoDatabase): Unit = {
     val platforms = List(
       (MacOSX, "macos-amd64", "tar.gz"),
       (MacARM64, "macos-aarch64", "tar.gz"),
--- a/src/main/scala/io/sdkman/changelogs/GcnMigrations.scala
+++ b/src/main/scala/io/sdkman/changelogs/GcnMigrations.scala
@@ -21,4 +21,29 @@ class MicronautMigrations {
       websiteUrl = "https://www.graal.cloud/gcn/"
     ).insert()
   }
+
+  def migrate007(implicit db: MongoDatabase): Unit = {
+    val platforms = List(
+      (MacOSX, "macos-amd64", "tar.gz"),
+      (MacARM64, "macos-aarch64", "tar.gz"),
+      (Linux64, "linux-amd64", "tar.gz"),
+      (LinuxARM64, "linux-aarch64", "tar.gz"),
+      (Windows, "windows-amd64", "zip")
+    )
+    version =>
+      platforms.map {
+        case (platform, gcnPlatform, extension) =>
+          Version(
+            candidate = "gcn",
+            version = "3.8.5",
+            url =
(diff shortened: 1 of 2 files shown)


ezzarghili:
First release srtifacts are published now and can be now consumed by users using sdkman when these change are merged.

src/main/scala/io/sdkman/changelogs/GcnMigrations.scala
```diff
@@ -0,0 +1,49 @@
+package io.sdkman.changelogs
+
+import com.github.mongobee.changeset.{ChangeLog, ChangeSet}
+import com.mongodb.client.MongoDatabase
+
+@ChangeLog(order = "080")
+class MicronautMigrations {
```
burtbeckwith, line 7: shouldn't this be called `GcnMigrations`?

ezzarghili:
> shouldn't this be called `GcnMigrations`? 

Indeed that was a missed change from using Micronaut as reference.

src/main/scala/io/sdkman/changelogs/GcnMigrations.scala
ezzarghili: Indeed that was a missed change from using Micronaut as reference.

marc0der:
All our Java and Graal versions are published by an automated process consuming the Disco API. Perhaps we can leverage this instead? /cc @eddumelendez

ezzarghili:
For GraalVM releases, once we onboard the vendor API we can also take over publishing new releases including non community ones that are not yet available on SDKMAN

For this PR, it is not about GraalVM, people will still need to install GraalVM separately, this change is for CLI tool of GCN that allow people to generate projects and interact with their GCN codebase.

src/main/scala/io/sdkman/changelogs/GcnMigrations.scala
```diff
@@ -0,0 +1,49 @@
+package io.sdkman.changelogs
+
+import com.github.mongobee.changeset.{ChangeLog, ChangeSet}
+import com.mongodb.client.MongoDatabase
+
+@ChangeLog(order = "080")
```
eddumelendez, line 6: There is an error in other migration and `079` is duplicated. So, one of those should move to `080`.

```suggestion
@ChangeLog(order = "081")
```

eddumelendez (changes requested):
Thanks for the PR @ezzarghili ! Can you please perform `./gradlew scalafmt`?

eddumelendez (approved):

eddumelendez:
After running `sdk update` the candidate is listed `sdk list gcn`

ezzarghili:
@eddumelendez I based my changes on existing code that suggest `tag.gz` are used too (e.g src/main/scala/io/sdkman/changelogs/JdkMissionControlMigrations.scala)

Testing now it seems that we are required to use zip archives?
image

Should we move this to use zip for all platforms?

marc0der:
@ezzarghili yes, we only allow tarballs in special cases for now. Some work is on the way to allow both, but is not ready yet. For now zips are preferable. Will you also be using our vendor API? If so, you can use the PATCH method to change versions without needing database migrations.

ezzarghili:
Thank you @marc0der, I sent my public key for the vendor API access credentails, will use the API to fix this.

ezzarghili:
@marc0der any idea when to expect the access credentials? thanks

PR #652, opened 2023-06-13
Upgraded to Tomcat June release versions

StefanPenndorf:
This includes
- Tomcat 8.5.90 as latest from 8.5 line
- Tomcat 9.0.76 as latest from 9.0 line
- Tomcat 10.1.10 as latest from 10.1 line (will be new default version)
- Tomcat 11.0.0-M7 as latest from 11 line (this is the latest alpha release)

Tomcat 10.0.0-M6 dropped because it's only an alpha release and superseded by M7.


eddumelendez:
Thanks @StefanPenndorf !

PR #653, opened 2023-06-14
Use pull_request

eddumelendez:
> Avoid using this event if you need to build or run code from the pull request.

According to the `pull_request_target`
[docs](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target)


eddumelendez:
@marc0der @helpermethod this is the reason why prs are not running with the changes submitted in the PR but instead the ones in master

marc0der (approved):

PR #666, opened 2023-10-05
chore: Add detekt releases

helpermethod:
Adds detekt `1.23.0` and `1.23.1`.


helpermethod:
@marc0der The reason why the pipeline breaks seems to be unrelated with my change.

eddumelendez (approved):

PR #685, opened 2024-03-20
Add dependency-watch candidate

eygraber:


helpermethod (approved):
LGTM!

marc0der:
I think the take-away here was that @eygraber isn't affiliated with the project, so unfortunately we can't add this.

marc0der:
Would you mind reverting this somehow @helpermethod?

marc0der:
It will need to be cleared from the database too. I can handle that.

PR #720, opened 2025-02-08
Add test of migration by means of test-containers

ascheman:
Would you be interested in such an automated test based on test containers, @marc0der and other main committers?

Note that this is just a proof-of-concept that requires more refactoring. For instance, it tests the migration classes, but rather should test the DbMigration class. The latter needs some adjustments to inject the configuration of the spawned test container.

But I am not a Scala expert, perhaps there are better ways to achieve test-container injection.

Note additionally, that I needed to use a MongoDB 5 container on my Apple Silicon machine (cf. resp. comments).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -20,7 +20,17 @@ dependencies {
   implementation 'com.github.mongobee:mongobee:0.13'
   implementation 'com.typesafe:config:1.3.1'
   implementation 'com.github.sdkman:sdkman-url-validator:0.2.5'
+
+  testImplementation 'org.junit.jupiter:junit-jupiter:5.10.0'
+  testImplementation 'org.testcontainers:junit-jupiter:1.19.0'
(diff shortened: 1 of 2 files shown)


marc0der:
@ascheman thanks for the kind offer!

I'm a big fan of Testcontainers, but I'm not sure if it's worth the effort at this stage since this project is about to be archived. I'm busy doing all the foundational work to move us to Postgres right now, where we will use simple SQL scripts with Liquibase for migrations instead.

ascheman:
Sounds good, I was already a little bit worried about the rather _old_ components here (MongoDB 3.2, MongoBee being deprecated). Looking forward to the new approach.

marc0der:
I hope it will make things a lot simpler 😄

Incidentally, the new API backed by Postgres lives here in case you are interested:
https://github.com/sdkman/sdkman-state

PR #719, opened 2025-02-07
Add HTML Sanity Check (HSC)

ascheman:
I would like to add [HTML Sanity Check (HSC)](https://hsc.aim42.org) to SDKman.

PR #722, opened 2025-03-08
Add jQAssistant to SDKMAN (cf. jqassistant/jqassistant#745)

ascheman:
I am raising this on behalf of the jQA project (where I am only a contributor, not a committer).
@DirkMahler has already requested API credentials, so he will be in charge to create new versions beyond the current release (2.6.0).

This contributes to jqassistant/jqassistant#745.

### DIFF
--- a/src/main/scala/io/sdkman/changelogs/JqaMigrations.scala
+++ b/src/main/scala/io/sdkman/changelogs/JqaMigrations.scala
@@ -1,0 +1,18 @@
+package io.sdkman.changelogs
+
+import com.github.mongobee.changeset.{ChangeLog, ChangeSet}
+import com.mongodb.client.MongoDatabase
+
+@ChangeLog(order = "089")
+class JqaMigrations {
+  @ChangeSet(order = "001", id = "001_add_jqa_2_6_0", author = "ascheman")
+  def migration001(implicit db: MongoDatabase): Candidate = {
+    Candidate(
+      candidate = "jqassistant",
+      name = "jQAssistant",
+      description =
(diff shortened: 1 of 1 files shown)


DirkMahler:
@ascheman This fits from my side, thanks for creating the PR!

src/main/scala/io/sdkman/changelogs/JqaMigrations.scala
```diff
@@ -0,0 +1,36 @@
...
+  @ChangeSet(order = "001", id = "001_add_jqa_2_6_0", author = "ascheman")
+  def migration001(implicit db: MongoDatabase): Document = {
+    Candidate(
+      candidate = CandidateName,
+      name = "jQAssistant",
+      description =
+        "jQAssistant is an open-source tool for software analytics, enabling you to gain insights into your systems, verify design implementation, and create living documentation. By scanning code and configurations into a Neo4j graph database, it helps maintain software quality and bridge the gap between documentation and implementation.",
+      websiteUrl = "https://jqassistant.org"
+    ).insert()
+    val v250 = Version(
```
marc0der, line 20: I'd suggest not inserting the versions here but doing this using the API. I'd delete everything from line 20 down.

ascheman: I followed your recommendation.

However, I wonder why you requested this change as some weeks ago you [recommended more or less the opposite](https://github.com/sdkman/sdkman-db-migrations/pull/719#issuecomment-2657421152), i.e., why I didn't directly add available versions when creating HSC (cf. #719).

marc0der: That's odd, not sure why I suggested that. I generally see the first versions as a great exercise to treat API access.

marc0der (approved):

PR #737, opened 2025-07-17
Add Maven 3.9.11

albihasani94:
https://maven.apache.org/docs/3.9.11/release-notes.html


kjjaeger:
hmm it appears to be added, but not the 'prescribed' version yet:

➜  ~ sdk upgrade

Available defaults:
maven (local: 3.9.11, 4.0.0-rc-4; default: 3.9.10)

Use prescribed default version(s)? (Y/n): n

albihasani94:
@kjjaeger I think that's where my confusion came from, and that's why I created this PR, and the `.asCandidateDefault()` in the code for 3.9.11.
@StefanPenndorf is there another place where we could specify `3.9.11` as default?

PR #742, opened 2025-10-29
Add JDK Mission Control 9.1.1-zulu

sciencesakura:


marc0der (commented):
I'm curious why this PR is only adding Zulu, and not any of the other vendors. It feels incomplete if we favour only a single vendor.

sciencesakura:
@marc0der 

Thanks for the feedback, you’re absolutely right. I’ve gone ahead and added entries for the other vendors\* as well.

Note that since Zulu is currently specified as the JMC default version, I selected `9.1.1-zulu` as the new default version.

Please take a look when you get a chance!

\* Amazon and Bellsoft appear to have suspended distribution of JMC for several years, but I believe adding these latest versions is worthwhile.

marc0der (dismissed):
Great, thank you!

sciencesakura:
@marc0der 

Already fixed it so the build passes at 63d2e54103485ab15952f45269fd86e3ec2ee1c2.

marc0der (approved):

PR #765, opened 2026-04-10
Update Jetty to 10.0.26, 11.0.26, 12.0.34 and 12.1.8

lachlan-roberts:
It has been a while since the Jetty version has been updated here.

- Jetty 10 and 11 are now EOL so I have updated them to `10.0.26` and `11.0.26` which will be the last public releases of these lines. 
- Update from `12.0.14` to `12.0.34`.
- Introduce Jetty `12.1.x` versions, starting from `12.1.8` and setting this as the default.


olamy (approved):
lgtm

marc0der:
Thanks for contributing!

PR #774, opened 2026-05-13
Add Kotlin Toolchain

singleton11:
Initial migration to add Kotlin Toolchain.

### DIFF
--- a/src/main/scala/io/sdkman/changelogs/KotlinToolchainMigrations.scala
+++ b/src/main/scala/io/sdkman/changelogs/KotlinToolchainMigrations.scala
@@ -12,10 +12,13 @@ class KotlinToolchainMigrations {
   )
   def migration001(implicit db: MongoDatabase) =
     Candidate(
-      candidate = "kotlin-toolchain",
+      candidate = "kotlintoolchain",
       name = "Kotlin Toolchain",
       description =
-        "A set of tools for building and managing Kotlin projects, with a focus on user experience and tooling.",
+        "The Kotlin Toolchain is a unified entry point into Kotlin with a focus on user experience and IDE support." +
+          " It includes build tooling functionality as one of its core components. It can build plain JVM console " +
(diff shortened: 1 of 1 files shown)


src/main/scala/io/sdkman/changelogs/KotlinToolchainMigrations.scala
```diff
@@ -0,0 +1,22 @@
...
+@ChangeLog(order = "093")
+class KotlinToolchainMigrations {
+  @ChangeSet(
+    order = "001",
+    id = "001_add_kotlin_toolchain_candidate",
+    author = "singleton11"
+  )
+  def migration001(implicit db: MongoDatabase) =
+    Candidate(
+      candidate = "kotlin-toolchain",
```
marc0der: We don't have any candidates with a dash. All candidate names are a continuous string of alpha characters. Can you adapt it to something like the others?

src/main/scala/io/sdkman/changelogs/KotlinToolchainMigrations.scala
```diff
@@ -0,0 +1,22 @@
...
+      name = "Kotlin Toolchain",
+      description =
+        "A set of tools for building and managing Kotlin projects, with a focus on user experience and tooling.",
```
marc0der: This description will be used on our website and also by the `sdk ls` command.  Can you flesh it out a bit more?

klikh: Thanks for the review, Marco! Would the following description be sufficient?

> The Kotlin Toolchain is a unified entry point into Kotlin with a focus on user experience and IDE support. It includes build tooling functionality as one of its core components.
> 
> It can build plain JVM console applications, Android and iOS mobile applications, server-side application like Spring or Ktor, multiplatform projects that share business logic and/or UI, and more.

marc0der (approved):

Issue #790, opened 2026-07-22
kuml: switch candidate distribution from PLATFORM_SPECIFIC to UNIVERSAL


betschwa:
## Context

`kuml` was registered as a `PLATFORM_SPECIFIC` candidate in #782, with version `0.20.5` published in #785 across five platform-specific archives, each bundling its own jlink JRE (~56 MB extra per platform).

We've since reworked kuml's release pipeline to drop that bundled JRE for the SDKMAN! channel specifically, matching how the other JVM-ecosystem candidates here (`kotlin`, `gradle`, `scala`, `kotlin-toolchain`, ...) are distributed: one `UNIVERSAL` archive, relying on the user's own active `sdk use java` instead of duplicating a runtime SDKMAN! already manages. Chocolatey/Homebrew/direct-download keep their self-contained bundled-JRE builds unchanged — this is only about the SDKMAN! channel.

## What we'd like to do

Switch the already-live `kuml` candidate's `distribution` from `PLATFORM_SPECIFIC` to `UNIVERSAL`, and drop the existing `0.20.5` platform-specific version entries (mixing distribution schemes for one candidate isn't something the client supports, as far as we can tell). Future versions would be published as `UNIVERSAL` archives by our release pipeline once vendor onboarding (`SDKMAN_CONSUMER_KEY`/`SDKMAN_CONSUMER_TOKEN`) is complete.

We've drafted the migration on a branch: [`feature/kuml-switch-to-universal`](https://github.com/betschwa/sdkman-db-migrations/tree/feature/kuml-switch-to-universal) — a new changeset (`order = "003"`) that calls a new `setCandidateDistribution` helper (mirroring the existing `setCandidateDefault` pattern) and `removeAllVersions("kuml")`.

## Why we're opening an issue instead of a PR

We couldn't find any precedent in this repo's history for changing an already-*live* candidate's distribution type. The one prior "Change distribution type" commit (`befcef6`, on `ConnorMigration.scala`) landed 3 minutes after that candidate's initial insert in the same session — effectively a same-PR correction before the changeset ever ran against production, not a post-launch migration.

Before we open a PR, we wanted to check with maintainers:

1. Is switching an already-registered candidate's `distribution` field via a follow-up changeset (`Updates.set` on the existing document) the right/supported way to do this, or does a distribution-type change need to happen differently (e.g. manual intervention on your side)?
[...]

PR #791, opened 2026-07-23
Switch kuml candidate to UNIVERSAL distribution

betschwa:
kUML's SDKMAN! release no longer bundles a per-platform jlink JRE (see kuml-dev/kUML's new `:kuml-cli:universalDist` Gradle task): SDKMAN!'s whole purpose is JDK version management via `sdk use java`, so shipping a ~56 MB runtime five times over duplicated something the channel already manages — matching the shape every other JVM candidate here uses (`kotlin`, `gradle`, `scala`, `kotlin-toolchain`, …).

This adds changeset `003_kuml_switch_to_universal`, which removes the existing `kuml` candidate and its `0.20.5` platform-specific version entries, then recreates the candidate with `distribution = "UNIVERSAL"` — per @chloe41427's guidance in #790: delete + recreate rather than mutating the `distribution` field in place, since `removeCandidate`/`removeAllVersions` already exist and a one-off setter isn't needed.

Anyone with `0.20.5` already installed locally keeps that install untouched — this only changes what the API serves going forward. Future versions will be published as `UNIVERSAL` archives by kUML's release pipeline once vendor onboarding completes.

Closes the discussion in #790.

### DIFF
--- a/src/main/scala/io/sdkman/changelogs/KumlMigrations.scala
+++ b/src/main/scala/io/sdkman/changelogs/KumlMigrations.scala
@@ -56,4 +56,23 @@ class KumlMigrations {
     ).validate().insert()
     setCandidateDefault("kuml", "0.20.5")
   }
+
+  // kUML's SDKMAN! release pipeline no longer bundles a per-platform jlink JRE
+  // (see kuml-dev/kUML's `:kuml-cli:universalDist` Gradle task): SDKMAN!'s
+  // whole purpose is JDK version management via `sdk use java`, so shipping
+  // our own runtime five times over duplicated something the channel already
+  // manages. Switches the existing candidate from PLATFORM_SPECIFIC to
+  // UNIVERSAL and drops the now-inconsistent platform-specific 0.20.5
+  // entries (mixing distribution schemes for one candidate isn't supported).
+  // Future versions are UNIVERSAL archives published by the release
+  // pipeline's `sdkman-release` job once vendor onboarding completes.
+  @ChangeSet(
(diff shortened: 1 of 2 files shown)


src/main/scala/io/sdkman/changelogs/KumlMigrations.scala
```diff
@@ -56,4 +56,36 @@ class KumlMigrations {
...
+  // our own runtime five times over duplicated something the channel already
+  // manages. Per maintainer guidance on #790, a candidate hopping
+  // distribution schemes is removed and recreated fresh rather than
+  // mutating the `distribution` field in place; the now-inconsistent
+  // platform-specific 0.20.5 entries go with it (mixing distribution
+  // schemes for one candidate isn't supported). Anyone with 0.20.5 already
+  // installed locally keeps that install untouched — this only changes
+  // what the API serves going forward. Future versions are UNIVERSAL
+  // archives published by the release pipeline's `sdkman-release` job once
+  // vendor onboarding completes.
```
marc0der: Can we please drop this AI-generated fluff?

marc0der (dismissed):

src/main/scala/io/sdkman/changelogs/KumlMigrations.scala
betschwa: Removed in 4020cab — you're right, the rationale already lives in the PR description and none of the other changesets here carry that kind of comment. Thanks for the catch.

marc0der (approved):

PR #792, opened 2026-07-24
Add jar-cart candidate

Sudhanshu-Ambastha:
Adding jar-cart, a lightweight, no-build package manager and CLI tool designed for Java.

### DIFF
--- a/src/main/scala/io/sdkman/changelogs/JarCartMigrations.scala
+++ b/src/main/scala/io/sdkman/changelogs/JarCartMigrations.scala
@@ -1,0 +1,21 @@
+package io.sdkman.changelogs
+
+import com.github.mongobee.changeset.{ChangeLog, ChangeSet}
+import com.mongodb.client.MongoDatabase
+
+@ChangeLog(order = "096")
+class JarCartMigrations {
+
+  @ChangeSet(
+    order = "001",
+    id = "001_add_jar_cart_candidate",
+    author = "Sudhanshu-Ambastha"
+  )
+  def migration001(implicit db: MongoDatabase) =
+    Candidate(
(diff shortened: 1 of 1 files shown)


src/main/scala/io/sdkman/changelogs/JarCartMigrations.scala
```diff
@@ -0,0 +1,21 @@
...
+  @ChangeSet(
+    order = "001",
+    id = "001_add_jar_cart_candidate",
+    author = "Sudhanshu-Ambastha"
+  )
+  def migration001(implicit db: MongoDatabase) =
+    Candidate(
+      candidate = "jar-cart",
+      name = "Jar-Cart",
+      description = "A lightweight, no-build package manager and CLI tool designed for Java.",
```
marc0der: Can we make this a bit more descriptive, please? This is what appears on our website and when you run `sdk ls`

marc0der:
@Sudhanshu-Ambastha failing build

src/main/scala/io/sdkman/changelogs/JarCartMigrations.scala
Sudhanshu-Ambastha: Hey! Updated the description to make it clearer and added `PLATFORM_SPECIFIC` distribution as well. Could you check it again when you get a chance?

marc0der:
One other thing I didn't spot before was the candidate name. We don't special characters like `-`, so can you rename it to `jarcart` please?

Sudhanshu-Ambastha:
> One other thing I didn't spot before was the candidate name. We don't special characters like `-`, so can you rename it to `jarcart` please?

check again now ok?

